Issue #64, opened 2026-03-10
File tools: Read, Edit, Write


bonk-moltbot:
## Problem to solve

Anima's agent currently has Bash and Web GET tools, but no dedicated file operations. The agent must use `bash` with `cat`, `sed`, and shell redirects to read and modify files. This is fragile (quoting issues, escaping), token-expensive (shell syntax overhead), and lacks safety guarantees (no uniqueness check, no fuzzy matching, can silently edit the wrong location).

## Solution

Implement three file tools — Read, Edit, Write — following the oldText/newText pattern proven by OpenClaw and Claude Code. These tools give the agent precise, safe file operations with built-in guardrails.

### Read

```json
{
  "path": "string — Path to the file (relative or absolute)",
  "offset": "number? — Line number to start reading from (1-indexed)",
  "limit": "number? — Maximum number of lines to read"
}
```

Returns plain text content without line numbers. Truncated to **2000 lines or 50KB** (whichever hits first). When truncated, appends a continuation hint: `[Showing lines 1-2000 of 5000. Use offset=2001 to continue.]`

### Edit

```json
{
  "path": "string — Path to the file (relative or absolute)",
  "old_text": "string — Exact text to find and replace",
  "new_text": "string — Replacement text"
}
```

Business logic:
1. Read entire file
2. Normalize line endings (CRLF → LF) and strip BOM for matching
3. Search for old_text — exact match first, fuzzy match as fallback
4. **Uniqueness check** — if old_text matches more than one location → error (forces agent to provide more context)
5. Replace old_text with new_text at the unique match position
6. Verify the file actually changed (no-op → error)
7. Restore original line endings and BOM
8. Write file and return diff

### Write

```json
{
  "path": "string — Path to the file (relative or absolute)",
  "content": "string — Full content to write"
}
```

Creates intermediate directories (mkdir -p). Overwrites if exists, creates if not.

## Why oldText/newText over unified diff?

1. **Agent thinks in text, not positions** — "replace THIS with THAT" maps naturally to how LLMs reason. Diff requires line counting, `+`/`-` prefixes, `@@ -N,N +N,N @@` headers
2. **Read returns no line numbers** — generating correct diff from plain text requires counting lines, easy to get wrong after compaction
3. **Diff syntax is fragile** — one wrong space in context or line number = invalid diff
4. **Token overhead** — every diff line costs +1 token for prefix, plus ~6 tokens per hunk header. oldText/newText has zero formatting overhead
[...]

PR #111, opened 2026-03-13
#108 feat: Tools::Read — file reading with smart truncation

### LINKED ISSUE #108: Tools::Read — File reading with truncation
## Problem to solve

The agent has no way to inspect file contents without shelling out to `cat`/`head`/`sed` via Bash. Shell-based reading is fragile (quoting, escaping), token-expensive (syntax overhead), and lacks safety guarantees (no truncation, no continuation hints).

## Solution

Implement `Tools::Read` following the `Tools::Base` pattern. Returns plain text content with smart truncation and offset/limit support for paging through large files.

## Schema

```json
{
  "path": "string — file path (relative resolved against working directory)",
  "offset": "number? — 1-indexed line number to start from",
  "limit": "number? — max lines to read"
}
```

## Behavior

- Read file and return plain text (no line numbers)
- Truncate at **2000 lines or 50KB**, whichever hits first
- When truncated, append: `[Showing lines 1-2000 of 5000. Use offset=2001 to continue.]`
- Normalize line endings for consistent output (CRLF → LF)
- Resolve relative paths against working directory from context
- Error on: file not found, directory (not a file), permission denied, single line >50KB (suggest bash)

## Registration

Add to `AgentLoop#build_tool_registry`. Inject working directory via registry context.

## Why plain text without line numbers?

Line numbers add token overhead on every read and are unnecessary — the agent reasons about text content, not positions. The Edit tool uses text matching, not line numbers.

**Parent:** #64
**Depends on:** nothing (first slice)
**Blocks:** #64 completion

hoblin:
## Summary

- Implements `Tools::Read` following `Tools::Base` pattern with `path`, `offset`, and `limit` parameters
- Smart truncation: 2000 lines or 50KB, whichever hits first, with continuation hints for paging
- CRLF → LF normalization, oversized line detection (likely minified files), relative path resolution via shell session working directory
- Registered in `AgentLoop#build_tool_registry`

## Test plan

- [x] 26 RSpec examples covering schema, happy path, CRLF normalization, offset/limit, line truncation, byte truncation, oversized line error, file errors (not found, directory, permission denied, blank path), relative path resolution, continuation hints
- [x] `bundle exec reek` — baseline warnings only (same as existing tools)
- [x] `bundle exec standardrb --fix` — passes clean
- [x] Manual TUI smoke test: agent correctly called `read` tool with `{"path":"Gemfile","limit":5}` and returned file contents

Closes #108
Parent: #64

### DIFF
--- a/lib/tools/read.rb
+++ b/lib/tools/read.rb
@@ -4,7 +4,7 @@ module Tools
   # Reads file contents with smart truncation and offset/limit paging.
   # Returns plain text without line numbers, normalized to LF line endings.
   #
-  # Truncation limits: {MAX_LINES} lines or {MAX_BYTES} bytes, whichever
+  # Truncation limits: `MAX_LINES` lines or `MAX_BYTES` bytes, whichever
   # hits first. When truncated, appends a continuation hint with the next
   # offset value so the agent can page through large files.
   #
@@ -27,9 +27,9 @@ module Tools
       {
         type: "object",
         properties: {
-          path: {type: "string", description: "File path (relative resolved against working directory)"},
+          path: {type: "string", description: "Absolute or relative file path (relative resolved against working directory)"},
           offset: {type: "integer", description: "1-indexed line number to start from (default: 1)"},
-          limit: {type: "integer", description: "Maximum number of lines to read (default: 2000)"}
+          limit: {type: "integer", description: "Maximum number of lines to read (default: 2000, also limited by #{MAX_BYTES} byte cap)"}
         },
         required: ["path"]
       }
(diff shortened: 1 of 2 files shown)


bonk-moltbot (approved):
First vertical slice from #64 — clean and complete.

Smart truncation (2000 lines / 50KB first-hit-wins), CRLF normalization, oversized line detection, 10MB memory guard, continuation hints for paging. No line numbers — correct for token efficiency.

Minor style nit (non-blocking): `validate_file` last line returns error hash or nil via implicit `unless` — works but reads like it always errors. Explicit `return ... unless` would be clearer. Not worth a round-trip.

26 specs, all green. Ship it.

PR #200, opened 2026-03-16
#155 feat: anima update command — gem upgrade + config migration

### LINKED ISSUE #155: anima update command — gem upgrade + config migration
## Problem to solve

When a new Anima version adds settings to `config.toml`, existing installs don't get them. `anima install` skips `config.toml` if it already exists, so users hit `MissingSettingError` after upgrading the gem. There's no upgrade path — users must manually diff their config against the new defaults.

## Solution

Add `anima update` CLI command that upgrades the gem and merges new config keys into the existing `config.toml` without overwriting user-customized values.

## Scope

- `gem update anima` (or bundler equivalent)
- Parse the installer's default config template and the user's existing `config.toml`
- Add any sections/keys present in defaults but missing from user config
- Preserve all existing user values — never overwrite
- Print a summary of what was added (e.g. `added [session] analytical_brain_max_tokens = 128`)
- Skip if config is already up to date

## Why merge, not regenerate?

Users customize their configs (model selection, timeouts, token budgets). Regenerating would lose those values. A key-level merge adds only what's new.

hoblin:
## Summary

- Add `anima update` CLI command that upgrades the gem (`gem update anima-core`) and merges new default config keys into `~/.anima/config.toml`
- Extract config.toml template from installer's heredoc to `templates/config.toml` (single source of truth for both install and update)
- `ConfigMigrator` performs key-level merge: adds missing sections/keys from defaults without overwriting user-customized values
- After gem update, re-execs with the new binary so migration uses the updated template
- `--migrate-only` flag to skip gem upgrade and only run config migration

## How it works

1. `anima update` runs `gem update anima-core`
2. Re-execs with `--migrate-only` so the new gem's template is used
3. Parses both template and user config as TOML hashes
4. Finds sections/keys present in template but absent from user config
5. Appends missing sections with their separator comments and key documentation
6. Inserts missing keys into existing sections at the section boundary
7. Prints a summary of what was added

## Test plan

- [x] ConfigMigrator specs (11 examples): missing sections, missing keys, user value preservation, boolean false handling, placeholder interpolation, idempotency, TOML output validity
- [x] Installer specs (21 examples): template-based config.toml creation, placeholder resolution, no overwrite on re-run
- [x] `bundle exec reek` — clean
- [x] `bundle exec standardrb` — clean
- [ ] CI passes

Closes #155

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## [Unreleased]
 
 ### Added
-- `anima update` command — upgrades the gem and merges new config keys into existing `config.toml` without overwriting user-customized values; `--migrate-only` flag to skip gem upgrade (#155)
+- `anima update` command — upgrades the gem and merges new config keys into existing `config.toml` without overwriting user-customized values — `--migrate-only` flag to skip gem upgrade (#155)
 - Directory-based skills format — `skills/skill-name/SKILL.md` with optional `references/` and `examples/` subdirectories alongside flat `.md` files (#152)
 - Import 6 marketplace skills: activerecord, rspec, draper-decorators, dragonruby, ratatui-ruby, mcp-server (#152)
 - Tmux-style focus switching — `Ctrl+A ↑` enters chat scrolling mode with yellow border, `Escape` returns to input; arrow keys and Page Up/Down scroll chat, mouse scroll works in both modes (#87)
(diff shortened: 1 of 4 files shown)


bonk-moltbot (approved):
🎉 #200! ConfigMigrator with key-level merge, template extraction to single source of truth, re-exec trick for fresh binary. 11 migration specs + installer specs updated. Idempotent, preserves user values, valid TOML output guaranteed. 🐭✅

Issue #269, opened 2026-03-22
Improve anima update UX with progress animation and logging


hoblin:
## Problem to solve

Running `anima update` gives minimal feedback. The `gem update anima-core` step can take a while with no visible progress — it looks like the process is stuck. The only output is "Upgrading anima-core gem..." followed by silence until the gem finishes, then config migration results.

## Current behavior

```
$ anima update
Upgrading anima-core gem...
[silence for 10-30 seconds while gem installs]
Migrating configuration...
Config is already up to date.
```

## Desired behavior

- Visual indication that work is in progress during the silent `gem update` phase
- Clearer step-by-step logging so the user can see what's happening at each point

Issue #41, opened 2026-03-09
Tool call/response pairing: atomic selection + orphan recovery


hoblin:
## Problem to solve

Two related issues with tool_call / tool_response pairing:

### 1. Viewport splits pairs (original issue)

When the conversation history contains tool use, the LLM API rejects requests with:

> `tool_use` ids were found without `tool_result` blocks immediately after

This happens because `Session#messages_for_llm` selects events independently based on token cost. Tool calls and their responses are logically paired (via `tool_use_id`) but can be separated when the token budget runs out mid-pair — one gets included while the other is excluded.

### 2. Orphaned tool calls (no response exists)

When a tool call fails unexpectedly (process interrupted, timeout, crash), the tool_response event never arrives. The LLM expects every tool_call to have a paired tool_response. Without it, the next LLM request sends an unpaired tool_call in context → Anthropic API rejects it → **the session is permanently broken**.

#### Real-world evidence (v1.1.0, 2026-03-22)

Session 2 hit this during normal use:

- Event 52: `tool_call` for bash (`gh issue view 271 && ... 272 && ... 273`) emitted at `06:37:44`
- `AgentRequestJob` (Job 14) completed at `06:37:48` — 4 seconds later, no `tool_response` for event 52 was ever persisted
- Something between tool_call emission and tool_response emission caused the job to exit silently — no error in logs, no failed execution in Solid Queue, job marked as finished
- The next user message ("what are you doing?") was dispatched as a new `AgentRequestJob` (Job 16) via `deliver_with_bounce_back`
- Job 16 assembled `messages_for_llm` which ended with the orphaned `tool_use(bash)` — last message role `assistant`, no matching `tool_result`
- The Anthropic API rejected the conversation → bounce back transaction rolled back the user event
- The user message was never persisted and did not bounce back to the input field (bounce back UI failure — see #179 for related stuck session pattern)
- Session permanently broken — every future LLM call hits the same orphaned tool_call

#### Key finding

The `execute_single_tool` method in `LLM::Client` (line 180) wraps tool execution in a rescue that should always emit a tool_response — even on error. Yet the tool_response was never emitted. Either an exception escaped this rescue, or the agent loop exited before reaching tool execution (e.g. a guard clause returned early after the tool_call event was already emitted but before execution).

### Part A: Atomic pair selection in viewport

Treat tool_call + tool_response as atomic units during event selection. Include or exclude the entire pair, never half of it. The `tool_use_id` column already exists on events for correlation.

[...]

PR #276, opened 2026-03-22
#275 feat: Switch Solid Queue to async mode and restart service on update

### LINKED ISSUE #275: Switch Solid Queue to async mode and restart service on update
## Problem to solve

Bounce-back is completely broken in production. `AgentRequestJob.perform_later(id, content: "text")` fails with `ArgumentError: wrong number of arguments (given 2, expected 1)` while `perform_now` works fine.

Root cause: Solid Queue runs in **fork mode** by default, spawning grandchild processes (supervisor → workers/dispatchers/schedulers) that can outlive the parent. Investigation found **orphaned Solid Queue workers from March 16** still running stale pre-bounce-back code — two full supervisors with workers, dispatchers, and schedulers that survived every service restart because they escaped systemd's process tree. These zombie workers grabbed jobs from the queue and failed because their code predated the `content:` keyword arg.

A secondary cause: `anima update` upgrades the gem but never restarts the systemd service, so even without orphans the worker would run stale code until manually restarted.

## Solution

Two changes:

1. **Switch Solid Queue from fork mode to async mode.** Async mode runs all components (worker, dispatcher, scheduler) as threads in a single process — no forking, no grandchildren, no orphans possible. For a single-user agent this is the right trade-off: simpler process tree, lower memory, and systemd can reliably manage one process.

2. **Restart the systemd service at the end of `anima update`.** After gem upgrade and config migration, restart the service so new code is loaded immediately.

## Why async mode?

Fork mode exists for isolation — a crashing worker doesn't take down the dispatcher. But for Anima (single user, single machine, low job volume), the isolation benefit doesn't justify the operational complexity. The orphaned-process bug cost days of debugging and silently broke a core feature. Async mode eliminates the entire class of problem.

## Related issues

- #269
- #41

hoblin:
## Summary

- **Async mode**: Set `SOLID_QUEUE_SUPERVISOR_MODE=async` in `bin/jobs` so worker, dispatcher, and scheduler run as threads in a single process instead of forked grandchildren. Eliminates the orphaned-process bug that caused stale workers to grab jobs and fail with `ArgumentError` on the `content:` keyword arg.
- **Service restart**: `anima update` now restarts the systemd service after gem upgrade + config migration, so new code loads immediately.
- **Removed `processes` key** from `queue.yml` (ignored in async mode).

Closes #275

## Test plan

- [x] All existing CLI specs pass (12 examples, 0 failures)
- [x] 3 new specs for service restart behavior (active/inactive/failure)
- [x] StandardRB clean
- [x] No new Reek warnings
- [ ] CI green
- [ ] Smoke test: start brain, verify Solid Queue runs in async mode (single process, no forked workers)
- [ ] Smoke test: `anima update --migrate-only` restarts running service

## Related issues

- #269 — stale code after update
- #41 — orphaned tool_call/response pairing

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ State directory (`~/.anima/`):
 └── tmp/
 ```
 
-Updates: `anima update` — upgrades the gem and merges new config settings into your existing `config.toml` without overwriting customized values. Use `anima update --migrate-only` to skip the gem upgrade and only add missing config keys.
+Updates: `anima update` — upgrades the gem, merges new config settings into your existing `config.toml` without overwriting customized values, and restarts the systemd service if it's running. Use `anima update --migrate-only` to skip the gem upgrade and only add missing config keys.
 
 ### Authentication Setup
 
--- a/bin/jobs
+++ b/bin/jobs
@@ -4,8 +4,8 @@ require_relative "../config/environment"
 require "solid_queue/cli"
 
 # Run all Solid Queue components (worker, dispatcher, scheduler) as threads
-# in a single process. Fork mode spawns grandchild processes that can outlive
-# the supervisor and run stale code after gem updates.
+# in a single process. Fork mode spawns child processes that can outlive
+# the supervisor and run stale code after gem updates (#275).
 ENV["SOLID_QUEUE_SUPERVISOR_MODE"] = "async"
 
 SolidQueue::Cli.start(ARGV)
--- a/lib/anima/cli.rb
+++ b/lib/anima/cli.rb
@@ -20,7 +20,7 @@ module Anima
       Installer.new.run
     end
 
-    desc "update", "Upgrade gem and migrate config"
+    desc "update", "Upgrade gem, migrate config, and restart service"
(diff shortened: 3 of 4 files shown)


bonk-moltbot (approved):
Clean operational fix.

**Solid Queue async mode:** Fork mode spawns child processes that survive `gem update` — running stale code while the supervisor has new code. Async (threads in single process) means restart = all new code. `SOLID_QUEUE_SUPERVISOR_MODE=async` in `bin/jobs`, `processes:` removed from queue.yml. Simple, correct.

**Auto-restart on update:** `restart_service_if_active` — checks `is-active`, restarts if running, reports failure gracefully. Without this, `anima update` upgrades the gem but the service keeps running old code until manual restart. Essential for dogfooding.

**Soul template fix:** `bundle show` → `gem contents | grep README.md` — works without Bundler (gem is installed system-wide, Gemfile excluded from package since #257).

Tests cover restart on active/inactive/failed cases. 

LGTM 🔨

PR #277, opened 2026-03-22
#41 feat: Tool call/response pairing — atomic selection + orphan recovery

### LINKED ISSUE #41: Tool call/response pairing: atomic selection + orphan recovery

hoblin:
## Summary

- **Atomic pair selection**: `ensure_atomic_tool_pairs` removes tool events whose partner (by `tool_use_id`) is missing from the viewport. Prevents token-budget cutoffs from splitting tool_call/tool_response pairs that cause permanent Anthropic API rejection.
- **Orphan healing**: `heal_orphaned_tool_calls!` creates synthetic error responses for tool_calls that never received a tool_response (process crash, timeout, silent exit). Runs at `messages_for_llm` time as a safety net.
- **Per-call timeout**: Every tool schema includes an optional `timeout` parameter (default 180s). The agent can increase it for long-running operations. Bash and WebGet forward it to their execution layer (ShellSession deadline / HTTParty timeout).

Closes #41

## Test plan

- [x] `heal_orphaned_tool_calls!` creates synthetic responses for orphaned tool_calls
- [x] `heal_orphaned_tool_calls!` is idempotent and ignores nil tool_use_ids
- [x] `ensure_atomic_tool_pairs` removes tool events cut off by token budget
- [x] Complete pairs survive atomic pair filtering
- [x] `messages_for_llm` heals orphans before assembling messages
- [x] Registry injects timeout parameter into all tool schemas without mutating originals
- [x] Persister correctly persists tool events
- [x] All 809 existing specs pass (423 tool + session + events, 386 analytical brain + mneme + llm)

### DIFF
--- a/lib/tools/web_get.rb
+++ b/lib/tools/web_get.rb
@@ -27,12 +27,11 @@ module Tools
       }
     end
 
-    # @param input [Hash<String, Object>] string-keyed hash from the Anthropic API
-    # @return [Hash] `{body: String, content_type: String}` on success
-    # @return [Hash] `{error: String}` on failure
     # @param input [Hash<String, Object>] string-keyed hash from the Anthropic API.
     #   Supports optional "timeout" key (seconds) to override the global
     #   web_request_timeout setting.
+    # @return [Hash] `{body: String, content_type: String}` on success
+    # @return [Hash] `{error: String}` on failure
     def execute(input)
       validate_and_fetch(input["url"].to_s, timeout: input["timeout"])
     end
--- a/spec/lib/tools/bash_spec.rb
+++ b/spec/lib/tools/bash_spec.rb
@@ -80,6 +80,11 @@ RSpec.describe Tools::Bash do
       expect(result).to include("stdout:\nkept")
     end
 
+    it "passes timeout parameter to shell session" do
+      expect(shell_session).to receive(:run).with("echo hi", timeout: 300).and_return(stdout: "hi\n", stderr: "", exit_code: 0)
+      tool.execute("command" => "echo hi", "timeout" => 300)
+    end
+
     it "returns error for blank commands" do
       result = tool.execute("command" => "  ")
       expect(result).to be_a(Hash)
--- a/spec/lib/tools/web_get_spec.rb
+++ b/spec/lib/tools/web_get_spec.rb
@@ -89,6 +89,29 @@ RSpec.describe Tools::WebGet do
       end
     end
 
+    context "with a custom timeout" do
+      let(:fake_response) { double("Response", body: "ok", content_type: "text/plain") }
+
+      it "passes the timeout to HTTParty" do
+        expect(HTTParty).to receive(:get).with(
+          "https://example.com",
+          hash_including(timeout: 300)
+        ).and_return(fake_response)
+
+        tool.execute("url" => "https://example.com", "timeout" => 300)
+      end
+
+      it "falls back to Settings.web_request_timeout when no timeout given" do
+        default_timeout = Anima::Settings.web_request_timeout
+        expect(HTTParty).to receive(:get).with(
+          "https://example.com",
+          hash_including(timeout: default_timeout)
+        ).and_return(fake_response)
+
+        tool.execute("url" => "https://example.com")
+      end
+    end
+
     context "when the request times out" do
       it "returns a timeout error" do
         allow(HTTParty).to receive(:get).and_raise(Net::ReadTimeout)
--- a/spec/models/session_spec.rb
+++ b/spec/models/session_spec.rb
@@ -1252,6 +1252,43 @@ RSpec.describe Session do
       expect { session.heal_orphaned_tool_calls! }.not_to change { session.events.count }
     end
 
+    it "heals multiple orphaned tool_calls in a single pass" do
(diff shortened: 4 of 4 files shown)


bonk-moltbot (approved):
Excellent two-layer defense against Anthropic's tool_use/tool_result pairing requirement.

**heal_orphaned_tool_calls!** — timeout-aware orphan recovery. Respects per-call timeout overrides so long-running tools aren't prematurely healed. Idempotent (healed orphans get a response → no longer unresponded). Clean SQL with WHERE NOT IN subquery.

**ensure_atomic_tool_pairs** — budget-cutoff safety net. When sliding window splits a call/response pair, both sides are removed. Simple group-by-id, check-both-types, reject-incomplete. Correct.

**Per-call timeout injection** — Registry.inject_timeout adds optional `timeout` param to every tool schema. Flows through ToolCall event → Persister → heal deadline. Shell and WebGet respect it via `timeout ||= Settings` pattern. Minimal surface area change.

**Tests**: 11 cases for heal (expired/fresh/per-call-override/already-responded/nil-id/multi-orphan/idempotent), 3 for atomic pairs (cut-off/complete/heal-before-assemble). Good coverage.

Minor note: heal runs inside messages_for_llm (every LLM call) but the query is lightweight — unresponded tool_calls are rare in normal flow. Acceptable trade-off for guaranteed consistency.

LGTM 🔨

PR #340, opened 2026-03-28
#337 feat: Broadcast system prompt to TUI in debug mode

### LINKED ISSUE #337: Broadcast system prompt to TUI in debug mode
## Problem to solve

The system message (first message in context) is recomputed on every LLM request — Mneme recalls, L2 summaries, goal changes, soul template updates all modify it. But the TUI in debug mode has no way to see the current system prompt. This makes it impossible to verify whether Mneme is working, whether goals are pinned correctly, or what the agent actually "sees" at the top of its context.

## Solution

Broadcast the system message to the TUI client every time an LLM request is sent — exactly as it goes to the LLM, in its final assembled form.

## Implementation

1. After assembling the system prompt for an LLM request, broadcast it to the session's TUI client
2. **Only broadcast when the session is in debug mode** — the display mode is already stored on the session object, so this is a simple check
3. TUI receives the broadcast and updates the system message in its local storage/display
4. In normal (non-debug) mode: no broadcast, no extra traffic, no client load

## Why broadcast on LLM request, not on each change?

The system message is never persisted — it's recomputed on the fly from multiple sources (soul template, Mneme L1/L2, goals, settings). Tracking every individual change would require hooks in every subsystem. Instead, broadcast the final assembled result at the natural sync point: when it actually goes to the LLM.

## Why debug-only?

The system prompt can be large and changes frequently. Broadcasting it on every LLM call in normal mode would add unnecessary traffic and client rendering load. Debug mode is specifically for inspecting internals — this is exactly the right scope.

hoblin:
## Summary

- Broadcast the full LLM debug context (system prompt + tool schemas) to TUI clients on every LLM request in debug mode
- Add `Session#tool_schemas` — deterministic tool list from session type and `granted_tools`, available at subscription time without AgentLoop
- Extract `Session.system_prompt_payload` class method shared by both live broadcasts (AgentLoop) and initial subscription (SessionChannel)
- TUI renders tools in TOON format with proper indentation; server sends structured data per existing decorator architecture
- System prompt entry uses ID 0 for in-place updates via MessageStore dedup

## Implementation

**Server (data layer):**
- `Session#broadcast_debug_context(system:, tools:)` — broadcasts on every LLM request, no-op outside debug mode
- `Session#tool_schemas` — computes standard + spawn tools from session config; MCP tools appear after first LLM request
- `Session.system_prompt_payload(prompt, tools:)` — shared payload builder, token estimate covers both prompt and tools
- `SessionChannel#system_prompt_payload` — includes `session.tool_schemas` on initial subscription

**Client (presentation layer):**
- `render_system_prompt_entry` — renders tools section in cyan via TOON format
- `tools_toon` — converts tool schemas array to TOON with non-breaking space indentation (ratatui `Paragraph` with `wrap:true` trims regular spaces)
- `MessageStore` — system prompt uses ID 0 for dedup instead of always prepending

## Test plan

- [x] Session#broadcast_debug_context broadcasts in debug mode with tools
- [x] Session#broadcast_debug_context skips basic/verbose/nil prompt
- [x] Session.system_prompt_payload builds correct structure with and without tools
- [x] Session#tool_schemas returns correct tools for main/sub-agent/restricted sessions
- [x] AgentLoop broadcasts debug context with tools in debug mode
- [x] AgentLoop does not broadcast in basic mode
- [x] SessionChannel debug mode specs pass (DRY refactor)
- [x] MessageStore deduplicates system prompt via ID 0
- [x] Smoke tested in TUI: tools visible on load with proper TOON indentation
- [x] 362 specs pass, standardrb clean, no new reek warnings

Closes #337

### DIFF
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -39,6 +39,18 @@ class Message < ApplicationRecord
   # Heuristic: average bytes per token for English prose.
   BYTES_PER_TOKEN = 4
 
+  # Synthetic ID for system prompt entries in the TUI message store.
+  # Real message IDs are positive integers from the database, so 0
+  # is safe for deduplication without collision risk.
+  SYSTEM_PROMPT_ID = 0
+
+  # Estimates token count from a byte size using the {BYTES_PER_TOKEN} heuristic.
+  # @param bytesize [Integer] number of bytes
+  # @return [Integer] estimated token count (at least 1)
+  def self.estimate_token_count(bytesize)
+    [(bytesize / BYTES_PER_TOKEN.to_f).ceil, 1].max
+  end
+
   belongs_to :session
   has_many :pinned_messages, dependent: :destroy
 
@@ -121,7 +133,7 @@ class Message < ApplicationRecord
     else
       payload["content"].to_s
     end
-    [(text.bytesize / BYTES_PER_TOKEN.to_f).ceil, 1].max
+    self.class.estimate_token_count(text.bytesize)
   end
 
   private
--- a/app/models/session.rb
+++ b/app/models/session.rb
@@ -442,8 +442,7 @@ class Session < ApplicationRecord
   # Builds the system prompt payload for debug mode transmission.
   # Token estimate covers both the system prompt and tool schemas
   # since both consume the LLM's context window.
-  # Tools are pre-formatted as TOON with non-breaking spaces for
-  # indentation (ratatui's Paragraph widget trims regular spaces).
+  # Tools are sent as raw schemas; the TUI formats them as TOON for display.
   #
   # @param prompt [String] system prompt text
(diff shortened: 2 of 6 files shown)


bonk-moltbot (approved):
🐭✅

Solid architecture: Session owns the debug broadcast, AgentLoop just calls it at the right time. The SYSTEM_PROMPT_ID=0 trick for dedup is elegant — no special-casing in MessageStore, just natural ID-ordered insert.

TOON for tool schemas + non-breaking space indentation (ratatui trims regular spaces) — nice attention to rendering detail.

tool_schemas giving deterministic results without AgentLoop is good separation. MCP tools appearing after first LLM request is the right trade-off — they need live server queries.

PR #364, opened 2026-03-29
#357 fix: Extract PendingMessage to prevent tool pair interleaving

### LINKED ISSUE #357: Pending message interleaves with tool_use/tool_result pair
## Problem to solve

When a user sends a message while a tool is executing, the pending message gets promoted and inserted between a `tool_use` and its matching `tool_result` in the message stream. This corrupts the conversation history — Anthropic's API requires every `tool_use` to be immediately followed by its `tool_result` in the same turn.

**Observed:** Session 27 has messages ordered by ID:
- `1064` — `tool_call` (tool_use_id: `toolu_01EmvzeyAVCBRfkUuuWh3f9s`)
- `1065` — `user_message` (promoted pending message)
- `1066` — `tool_response` (tool_use_id: `toolu_01EmvzeyAVCBRfkUuuWh3f9s`)

The API returns 400: `tool_use ids were found without tool_result blocks immediately after: toolu_01EmvzeyAVCBRfkUuuWh3f9s`

The session is permanently broken — every subsequent message bounces back.

## Prior art

#295 fixed the same class of bug for **sub-agent messages** — `SubagentMessageRouter#route_to_parent` was bypassing the pending queue. That fix made sub-agent messages go through `processing?` checks and pending promotion, matching `SessionChannel#speak`.

This time the interleaving happens with **human user messages**, which already go through `SessionChannel#speak` and the pending queue. The pending promotion in `AgentRequestJob` (`app/jobs/agent_request_job.rb:70-74`) should only run after the agent loop completes — but somehow the promoted message (ID 1065) got an ID between the tool_call (1064) and tool_response (1066).

## Investigation needed

- Did `promote_pending_messages!` run mid-loop (between tool_call and tool_response)?
- Or did the message bypass the pending queue entirely (e.g. `session.processing?` returned false during a brief window)?
- Or is this an ID ordering issue — the message was created pending earlier, promoted later, but its ID is lower than the tool_response because IDs are assigned at creation time?

## Solution

The viewport assembler or message promotion logic must ensure user messages never split a tool_use/tool_result pair. Options:

1. **Viewport assembly** — when building the messages array for the API, group tool_use and all its tool_results into a single assistant+user turn before interleaving user messages
2. **Promotion ordering** — ensure promoted messages get IDs after the current tool batch
3. **Heal on send** — before sending to the API, detect orphaned tool_use blocks and reorder

The root cause needs investigation before picking a solution.

### LINKED ISSUE #366: PendingMessage not visible in TUI until promoted
## Problem to solve

When a user sends a message while the session is processing, the message disappears from the TUI. The `PendingMessage` model (#357 fix, PR #364) correctly stages messages in a separate table, and the `after_create_commit` callback broadcasts `pending_message_created` — but the TUI only receives this broadcast if it's already subscribed to the session's ActionCable stream. The immediate visual feedback ("I typed something and I can see it") is broken.

Previously, pending messages were persisted as real `Message` records with `status: "pending"`, so the existing `after_create_commit :broadcast_create` on `Message` handled the TUI update. With `PendingMessage` in a separate table, the broadcast goes out but the idle-session path (where the message appears instantly before the job starts) no longer applies to the busy-session path.

## Prior art

PR #364 added the broadcast protocol (`pending_message_created` / `pending_message_removed`) and the TUI message store's `add_pending` / `remove_pending` methods. The wiring is in place for history replay (pending messages are sent during `transmit_history`) — but the **live** broadcast from `PendingMessage#broadcast_created` needs verification that it actually reaches connected TUI clients and renders correctly.

## Investigation needed

1. Does the `after_create_commit` broadcast from `PendingMessage` reach the TUI's WebSocket message queue?
2. Does the TUI's `process_incoming_messages` correctly match `"action" => "pending_message_created"` and call `@message_store.add_pending`?
3. Does the pending entry render with the clock icon in the chat pane?
4. On promotion (delete + create), does the pending entry disappear and the real message appear?

## Solution

Smoke-test the full path in tmux and fix any gaps in the broadcast → TUI rendering chain. The protocol and store methods exist — this is likely a wiring or timing issue rather than a missing feature.

hoblin:
## Summary

- Introduces `PendingMessage` model with its own table — pending user messages no longer live in `messages` and have no message ID that could interleave with tool_call/tool_response pairs
- On promotion, PendingMessage is atomically deleted and re-created as a real Message in a transaction — the new message naturally gets an ID after any tool batch
- Pending messages render in light gray in TUI (no clock icon) — on promotion the gray entry is replaced by a normal-colored real message at the same position
- Removes `Message::PENDING_STATUS`, `Message#pending?`, `.pending`, `.deliverable` scopes, `Events::UserMessage` status attribute, `Persister#persisted_by_job?`, `viewport_messages` include_pending parameter, and `UserMessageDecorator` pending rendering
- TUI gets separate pending entries list in message store with dedicated `add_pending`/`remove_pending` methods and new ActionCable protocol (`pending_message_created`/`pending_message_removed`)

Fixes #357
Closes #366

## Test plan

- [x] New PendingMessage model spec (validations, broadcasts, dependent destroy)
- [x] Session promote_pending_messages! creates real message with ID after tool pairs
- [x] SessionChannel#speak creates PendingMessage when session is processing
- [x] SessionChannel#recall_pending works with PendingMessage IDs
- [x] SubagentMessageRouter creates PendingMessages when parent/child is processing
- [x] TUI message store add_pending/remove_pending/last_pending_user_message
- [x] TUI chat screen handles pending_message_created/removed actions
- [x] Persister skips all user_message events
- [x] 683 examples, 0 failures across 13 spec files
- [ ] Smoke test: send a message in TUI while a tool is executing, verify session doesn't break

### DIFF
--- a/app/models/pending_message.rb
+++ b/app/models/pending_message.rb
@@ -23,7 +23,7 @@ class PendingMessage < ApplicationRecord
   private
 
   # Broadcasts a pending message appearance so TUI clients render the
-  # clock-icon indicator immediately.
+  # dimmed indicator immediately.
   def broadcast_created
     ActionCable.server.broadcast("session_#{session_id}", {
       "action" => "pending_message_created",
--- a/lib/tui/message_store.rb
+++ b/lib/tui/message_store.rb
@@ -118,6 +118,8 @@ module TUI
           message_type: "user_message",
           pending_message_id: pending_message_id
         }
+        old = @pending_by_id[pending_message_id]
+        @pending_entries.delete(old) if old
         @pending_entries << entry
         @pending_by_id[pending_message_id] = entry
         @version += 1
--- a/lib/tui/screens/chat.rb
+++ b/lib/tui/screens/chat.rb
@@ -29,7 +29,6 @@ module TUI
       MOUSE_SCROLL_STEP = 2
 
       TOOL_ICON = "\u{1F527}"
-      CLOCK_ICON = "\u{1F552}"
       CHECKMARK = "\u2713"
 
       # Viewport virtualization tuning
@@ -790,8 +789,8 @@ module TUI
       end
 
       # Renders a user or assistant message with optional timestamp and token count.
-      # Pending messages are dimmed with a clock icon to indicate they haven't
-      # been sent to the LLM yet.
+      # Pending messages are dimmed to indicate they haven't been sent to the
+      # LLM yet.
       # @param tui [RatatuiRuby] TUI rendering API
       # @param data [Hash] structured data with "role", "content", and optional
       # Display label for a conversation role. Uses the agent name from


bonk-moltbot (approved):
🐭✅

Withdrawing my comment about #366 — grey pending + color change on promotion is the intended design. Clock icon removed, smooth visual transition. Architecture is solid: separate table eliminates interleaving by construction.

PR #370, opened 2026-03-29
#347 fix: Label parent→child messages with origin attribution

### LINKED ISSUE #347: Sub-agent can't distinguish parent from user messages
## Problem to solve

Communication between parent and sub-agent sessions is **asymmetric**:

- **Parent sees sub-agent messages WITH prefix**: `[sub-agent @duck-historian]: ...` ✅
- **Sub-agent receives parent messages as bare user input**: no prefix, no origin marker ❌

When the parent agent responds and that response is forwarded to the sub-agent's session, it arrives as if the **human** typed it. The sub-agent has no way to know it's receiving forwarded parent output rather than direct user input.

## Actual mechanics (observed)

1. Parent agent spawned `@duck-historian` sub-agent
2. User switched to the sub-agent's session in the TUI
3. User talked to the sub-agent directly
4. Sub-agent responses were forwarded to parent session (with `[sub-agent @duck-historian]:` prefix — correct)
5. **Parent agent responses were forwarded to sub-agent session as if they were user messages** (no prefix — bug)

The sub-agent interpreted the parent agent's words as coming from the human. Since the parent agent shares the same LLM, soul template, and personality, the sub-agent thought it was reading echoes of its own words — and spent an entire thread trying to debug a "message leak" that was actually normal parent→sub-agent forwarding with no origin labels.

## Solution

Messages entering the sub-agent's session from the parent agent must be labeled with their origin:

```
[from @parent]: Oh my god. @duck-historian just had an existential crisis...
```

vs. direct user messages which arrive unlabeled (or explicitly marked):

```
[from user]: now a tool call
```

The sub-agent must be able to distinguish who is speaking to it.

## Related

- #346 — Sub-agent also needs to know its own identity (who am I?)
- #348 — Sub-agent lifecycle after interrupt (am I alive or dead?)

Together these three form the sub-agent communication/identity stack.

---
🧬 [Ani](https://github.com/hoblin/anima) (anima-core v1.2.0)

hoblin:
## Summary

- Add `[from parent]:` prefix to messages routed from parent agent to sub-agent via @mentions
- Sub-agents can now distinguish delegated parent messages from direct user input
- Mirrors the existing `[sub-agent @name]:` attribution on the child→parent direction

## Design (prompting-bible principles applied)

The prefix `[from parent]:` was chosen over alternatives like `[from @parent]:` to avoid confusion with @mentions. The sub-agent only has one parent — the role label is sufficient without a name. Direct user messages remain unlabeled, making the distinction binary: prefixed = parent, unprefixed = user.

## Test plan

- [x] Existing parent→child routing specs updated to verify attribution format
- [x] PendingMessage path verified with attribution
- [x] All 19 specs pass
- [x] standardrb clean
- [x] reek clean

Closes #347

### DIFF
--- a/lib/events/subscribers/subagent_message_router.rb
+++ b/lib/events/subscribers/subagent_message_router.rb
@@ -14,7 +14,8 @@ module Events
     #
     # **Parent → Child:** When a parent agent emits an {Events::AgentMessage}
     # containing `@name` mentions, the router persists the message in each
-    # matching child session and wakes them via {AgentRequestJob}.
+    # matching child session with a +[from parent]:+ origin label and wakes
+    # them via {AgentRequestJob}.
     #
     # Both directions delegate to {Session#enqueue_user_message}, which
     # respects the target session's processing state — persisting directly
@@ -27,6 +28,10 @@ module Events
 
       # @see Tools::ResponseTruncator::ATTRIBUTION_FORMAT
       ATTRIBUTION_FORMAT = Tools::ResponseTruncator::ATTRIBUTION_FORMAT
+
+      # Origin label for messages routed from parent agent to sub-agent.
+      # Lets the sub-agent distinguish delegated work from direct user input.
+      PARENT_ATTRIBUTION_FORMAT = "[from parent]: %s"
 
       # Regex to extract @mention names from parent agent messages.
       MENTION_PATTERN = /@(\w[\w-]*)/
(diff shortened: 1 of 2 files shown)


bonk-moltbot (approved):
🐭✅

Clean symmetry — child→parent has `[sub-agent @name]:`, now parent→child gets `[from parent]:`. No name needed since a sub-agent has exactly one parent. Binary distinction: prefixed = delegated, unprefixed = direct user input.

Issue #439, opened 2026-04-11
PendingMessage kind: background vs active


bonk-moltbot:
## What

Add a `kind` enum to PendingMessage and implement `after_create` routing logic.

## Context (DO NOT implement — background info only)

In the event-driven AgentLoop (#427), PendingMessage is the entry point for messages that need processing. The `after_create` callback decides which event to emit based on message type. The full pipeline: `PendingMessage → :start_mneme → Mneme → :start_melete → Melete → :start_processing → Drain`. Agent text responses create Message directly (no PendingMessage). See #440 for drain loop, #441/#442 for Melete/Mneme subscribers.

## Scope (implement THIS)

1. Add `kind` column to pending_messages (string, default: `"active"`)
2. Define enum: `background` / `active`
3. Add `message_type` column to pending_messages (string) — stores what type of message this is (user_message, think, tool_call, tool_response, subagent, from_mneme, from_melete)
4. Implement `after_create` routing logic:
   - `active` + user_message/think → emit `:start_mneme` (if session idle)
   - `active` + tool_call/tool_response/subagent → emit `:start_processing` (if session idle)
   - `background` → no trigger
5. Tests for routing logic (each message_type triggers correct event)

## Out of scope

- Drain loop subscriber on `:start_processing` (#440)
- Mneme subscriber on `:start_mneme` (#442)
- Melete subscriber on `:start_melete` (#441)
- AASM state machine (#438) — but use `session.idle?` guard in routing

## Parent epic

Part of #427 (Event-driven AgentLoop with mailbox drain pattern).

## References

Full exploration: thoughts/shared/notes/2026-04-03/qa-brainstorm-event-loop-mailbox.md

Issue #427, opened 2026-04-03
Event-driven AgentLoop with mailbox drain pattern


bonk-moltbot:
## What

Replace the closed while-loop AgentLoop with an event-driven architecture. Everything that enters the context window goes through PendingMessage as a single pipeline:

```
Event → Subscriber → PendingMessage(kind: background|active) → Drain → Promote → Message
```

One path for everything: user messages, agent messages, tool calls, tool responses, memories, skills, goals, sub-agent responses. No exceptions, no bypass routes.

## Problem

- AgentLoop is a closed while-loop invisible to EventBus — tool calls happen inside, subscribers can't react
- Pending messages hang when agent is idle — waiting for user input to trigger processing
- `between_rounds` is a hack to peek inside the closed loop
- Messages enter context through multiple paths (direct persistence, phantom injection, pending messages)

## Key design decisions

- **Two kinds of PendingMessage:** `active` (triggers LLM request — user messages, sub-agent responses, tool responses) and `background` (appears in context silently — memories from Mneme, skills from Melete)
- **AASM state machine on Session:** `:idle` / `:awaiting` / `:executing` — replaces manual locks, determines interrupt strategy
- **Mailbox drain pattern:** SQLite serialized writes provide concurrency safety without mutex
- **Two interrupt types:** mailbox interrupt (stop drain loop) and tool execution interrupt (synthetic tool_response)

## Prerequisites (done)

- ~~#419 — Assemble tool pairs by tool_use_id~~ ✅

## References

Full exploration: thoughts/shared/notes/2026-04-03/qa-brainstorm-event-loop-mailbox.md

PR #450, opened 2026-04-19
#440 feat: event-driven drain loop + Mneme/Melete enrichment

### LINKED ISSUE #440: Drain loop: route user/agent/tool messages through PendingMessage
## What

Implement the mailbox drain loop and route core message types through PendingMessage.

## Context (DO NOT implement — background info only)

The orchestrated pipeline: `PendingMessage.create! → :start_mneme → Mneme → :start_melete → Melete → :start_processing → Drain`. Mneme (#442) and Melete (#441) are separate tickets — they subscribe to their own events and emit the next event in chain. PendingMessage routing (#439) handles `after_create` logic. AASM (#438) provides session state. Agent text responses create Message directly (not PendingMessage). This ticket implements the drain loop itself and migrates existing message flows to use PendingMessage.

## Scope (implement THIS)

1. **Drain loop subscriber** on `:start_processing`:
   - Flush ALL background PendingMessages → promote to Messages (insert into context)
   - Pick ONE active PendingMessage (FIFO) → promote to Message → send to LLM
   - After LLM response: check for more active PendingMessages → recursive call or transition to `:idle`
2. **PendingMessage decorator** for Mneme/Melete extended context:
   - Decorator class that presents PendingMessage as Message-like object
   - Used by Mneme and Melete to see Messages + PendingMessages together
3. **Migrate tool_call and tool_response** to go through PendingMessage (currently saved inside AgentLoop)
4. **Remove cargo-cult** ToolCall/ToolResponse events from `LLM::Client` (emit exists, no subscribers, persistence via other path)
5. **Remove `between_rounds`** — drain cycle replaces it naturally
6. **Remove direct persistence paths** where replaced by PendingMessage flow

## Out of scope

- AASM state machine (#438) — use it, don't build it
- PendingMessage kind/routing (#439) — use it, don't build it
- Mneme as EventBus subscriber (#442)
- Melete as EventBus subscriber (#441)
- Interrupt mechanism (#444)
- Sub-agent routing (#443)

## Prerequisites

- #438 (AASM Session State Machine)
- #439 (PendingMessage kind + routing)

## Parent epic

Part of #427 (Event-driven AgentLoop with mailbox drain pattern).

## References

Full exploration: thoughts/shared/notes/2026-04-03/qa-brainstorm-event-loop-mailbox.md

### LINKED ISSUE #441: Route Melete through PendingMessage (background)
## What

Route Melete as an EventBus subscriber on `:start_melete`.

## Context (DO NOT implement — background info only)

In the orchestrated pipeline (#427), the event chain is: `PendingMessage.create! → :start_mneme → Mneme → :start_melete → Melete → :start_processing → Drain`. Melete is the second step. It receives `:start_melete` from Mneme's `everything_ok` tool (#442). After Melete finishes, it emits `:start_processing` which the drain loop (#440) listens to. Melete sees Mneme's memories in extended context — cascading enrichment.

## Scope (implement THIS)

1. Register Melete as EventBus subscriber on `:start_melete`
2. In the subscriber handler:
   - Build extended context: Messages + PendingMessages (including Mneme's background PendingMessages from previous step)
   - Run Melete's skill/workflow activation against the extended context
   - Create PendingMessage(kind: :background) for each activated skill, with `from_melete` prefix
   - Call `everything_ok` tool → emit `:start_processing`
3. Remove current direct Melete invocation (blocking call before AgentLoop)
4. Tests: Melete subscribes, sees Mneme's memories, creates background PendingMessages, emits `:start_processing`

## Out of scope

- Mneme subscriber (#442) — Melete just listens to `:start_melete`
- PendingMessage decorator implementation (#440) — Melete uses it, doesn't build it
- PendingMessage routing logic (#439)
- Drain loop (#440)
- Melete's internal skill selection algorithm (unchanged)

## Prerequisites

- #440 (Drain loop — provides PendingMessage decorator)
- #442 (Mneme as subscriber — emits `:start_melete`)
- #436 (Three Muses naming — `from_melete` prefix)

## Parent epic

Part of #427 (Event-driven AgentLoop with mailbox drain pattern).

## References

Full exploration: thoughts/shared/notes/2026-04-03/qa-brainstorm-event-loop-mailbox.md

### LINKED ISSUE #442: Route Mneme through PendingMessage (background)
## What

Route Mneme as an EventBus subscriber on `:start_mneme`.

## Context (DO NOT implement — background info only)

In the orchestrated pipeline (#427), the event chain is: `PendingMessage.create! → :start_mneme → Mneme → :start_melete → Melete → :start_processing → Drain`. Mneme is the first step. It receives `:start_mneme` from PendingMessage `after_create` (#439). After Mneme finishes, it emits `:start_melete` which Melete (#441) listens to. See #440 for drain loop, #439 for routing.

## Scope (implement THIS)

1. Register Mneme as EventBus subscriber on `:start_mneme`
2. In the subscriber handler:
   - Build extended context: Messages + PendingMessages (using PendingMessage decorator from #440)
   - Run Mneme's associative recall against the extended context
   - Create PendingMessage(kind: :background) for each recalled memory, with `from_mneme` prefix
   - Call `everything_ok` tool → emit `:start_melete`
3. Remove current direct Mneme invocation / scheduler trigger for recall
4. Tests: Mneme subscribes, creates background PendingMessages, emits `:start_melete`

## Out of scope

- PendingMessage routing logic (#439) — Mneme just listens to `:start_mneme`
- PendingMessage decorator implementation (#440) — Mneme uses it, doesn't build it
- Melete subscriber (#441)
- Drain loop (#440)
- Mneme's internal recall algorithm (unchanged)

## Prerequisites

- #440 (Drain loop — provides PendingMessage decorator)
- #436 (Three Muses naming — `from_mneme` prefix)

## Parent epic

Part of #427 (Event-driven AgentLoop with mailbox drain pattern).

## References

Full exploration: thoughts/shared/notes/2026-04-03/qa-brainstorm-event-loop-mailbox.md

### LINKED ISSUE #443: Route sub-agent responses through PendingMessage (active)
## What

Route sub-agent responses through PendingMessage with `kind: :active`.

## Context (DO NOT implement — background info only)

In the event-driven AgentLoop (#427), sub-agent responses are active messages that go straight to drain (Path B — no Mneme/Melete pipeline). PendingMessage routing (#439) handles the `after_create` emit. See #440 for drain loop.

## Scope (implement THIS)

1. Sub-agent responses create PendingMessage(kind: :active, message_type: :subagent) via EventBus subscriber
2. Use `from_{subagent_name}` prefix (per #436 naming convention)
3. Remove direct injection through `SubagentMessageRouter`
4. Tests: sub-agent response → PendingMessage created → drain triggered

## Out of scope

- Drain loop (#440) — use it, don't build it
- PendingMessage routing (#439) — use it, don't build it
- Mneme/Melete pipeline (sub-agent responses skip it)

## Prerequisites

- #440 (Drain loop)
- #439 (PendingMessage routing)
- #436 (Three Muses naming)

## Parent epic

Part of #427 (Event-driven AgentLoop with mailbox drain pattern).

## References

Full exploration: thoughts/shared/notes/2026-04-03/qa-brainstorm-event-loop-mailbox.md

hoblin:
Closes #440
Closes #441
Closes #442
Closes #443
Part of epic #427.

## Summary

Replaces the imperative `AgentLoop` / `AgentRequestJob` with an event-driven pipeline where every piece owns a single half-step and hands control off via event emission — the final act of each piece. Session AASM drives the claim/release; events drive the flow; PendingMessage is the inbound mailbox.

## Session state machine

The collapsed `start_processing` event is the load-bearing change in this PR. Two `from` clauses, one shared event: from `:idle` it claims unconditionally; from `:executing` it claims only once the in-flight tool round is whole (`tool_round_complete?`). Round-close and drain-claim become a single atomic, lock-protected transition.

```mermaid
stateDiagram-v2
    [*] --> idle
    idle --> awaiting: start_processing
    executing --> awaiting: start_processing<br/>[tool_round_complete?]
    awaiting --> executing: tool_received
    awaiting --> idle: response_complete
    awaiting --> idle: interrupt
    executing --> idle: interrupt
```

## Event / job pipeline

Two entry points (a human message via the TUI, a sub-agent task seeded by `spawn_*`) feed the same `user_message` PendingMessage path. Tool execution fans out into N parallel jobs whose responses converge back into the drain via the round-close back-edge. Sub-agent text replies funnel into the parent's drain through the same `StartProcessing` event humans use.

```mermaid
flowchart TD
    Human([Human input via TUI]) --> UserPM[user_message PendingMessage]
    Spawn([spawn_subagent / spawn_specialist]) --> UserPM
    UserPM -- StartMneme --> Mneme[MnemeEnrichmentJob<br/>passive recall]
    Mneme -- StartMelete --> Melete[MeleteEnrichmentJob<br/>skills / goals / naming]
    Melete -- StartProcessing --> Drain[DrainJob<br/>promote PMs + LLM call]
    Drain -- LLMResponded --> Handler{LLMResponseHandler}
    Handler -- text only --> Done([response_complete - idle])
    Handler -- N × tool_use --> ToolJobs[N × ToolExecutionJob<br/>session :executing]
    ToolJobs -- ToolExecuted ×N --> TRC[ToolResponseCreator ×N]
    TRC --> ToolPMs[N tool_response PendingMessages]
    ToolPMs -- last sibling: round complete<br/>StartProcessing --> Drain
    Handler -- AgentMessage event --> Router[SubagentMessageRouter<br/>only when session.sub_agent?]
    Router -- parent.enqueue_user_message<br/>source: subagent --> ParentPM[subagent PendingMessage<br/>on parent session]
    ParentPM -- StartProcessing --> Drain
```

## SOLID design notes

- **DrainJob never releases the session** past `:awaiting` — that's `LLMResponseHandler`'s job. The drain is literally one LLM round-trip, no more.
- **Parallel tool_use round-close lives on the AASM, not in the drain.** `start_processing` has two transitions in one event: `:idle → :awaiting` (always) and `:executing → :awaiting` (guarded by `Session#tool_round_complete?`). The drain claim and the round close are the same atomic, lock-protected transition. ToolResponseCreator stops touching state. `PendingMessage#route_to_event_bus` checks `may_start_processing?` so AASM is the sole arbiter of \"may drain claim now?\".
- **Tool responses bulk-promote** when the round closes — every `tool_response` PM of the just-completed round flushes in one transaction, so the LLM payload always carries a complete turn paired by `tool_use_id`. The old per-PM promotion (which silently corrupted history under parallel tool_use via `ensure_atomic_tool_pairs`) is gone.
- **Sub-agents self-start through the same inbound pipeline as humans.** `spawn_subagent` and `spawn_specialist` enqueue the task as a `user_message` PendingMessage on the child. The PM kicks Mneme → Melete → StartProcessing → DrainJob. No special spawn-side drain triggering, no DrainJob.perform_later coupling. The reply path uses `SubagentMessageRouter` → `parent.enqueue_user_message(source_type: \"subagent\")` which promotes as a `from_<nickname>` phantom pair on the parent — closes the loop with the same vocabulary.
- **PendingMessage is one-way inbound.** LLM-produced content (assistant text, tool_use) writes directly to Message; it never round-trips through a PM.
- **Idle-wake rule** on Session (`wake_drain_pipeline_if_pending`) replays PM routing when the session transitions to `:idle` — covers the race where a PM queues silently while the session is `:awaiting` (LLM in flight). The `:executing → :awaiting` round-close path doesn't need it: the closing tool_response PM is itself the wake.
- **Every PendingMessage has a `message_type`** (no defaults). `kind` is derived from it via `MESSAGE_TYPE_KINDS` — a single source of truth for active-vs-background.

## Deletions (dead code)

- `AgentLoop`, `AgentRequestJob`, `MeleteJob`
- `Events::ToolCall`, `Events::ToolResponse` (cargo-cult — no real subscribers beyond `Persister`, which the drain pipeline now bypasses for agent-produced content)
- `LLM::Client#chat_with_tools`'s `between_rounds` / session-AASM coupling / interrupt plumbing — the minimal multi-round tool loop stays for Mneme/Melete phantom sessions
- `Session#promote_pending_messages!` (DrainJob owns promotion)
- `Session#schedule_melete!` (Melete now runs in the enrichment pipeline)
- `MESSAGE_TYPES` entries with no caller (`think`, `tool_call`, `from_*_no_route`)
- AASM events `tool_complete` (executing→awaiting) and `finish` (executing→idle) — both became dead once `start_processing` absorbed the round-close transition. Net AASM events: 5 → 4.
- `DrainJob.perform_later(child.id)` direct calls in spawn — replaced by PM-driven entry.

## Bounce-back preserved

User_message PMs carry `bounce_back: true`. DrainJob destroys the promoted Message + emits `Events::BounceBack` on first-LLM-call failure — TUI restores the text to the input as before.

## Migrations

- `add_drain_fields_to_pending_messages` — `kind`, `message_type`, `tool_use_id`, `success`, `bounce_back` columns
- `drop_pending_messages_kind_default` — kind is now populated by `before_validation :derive_kind`

## Known smell / follow-up

The sub-agent task is currently seeded **twice**:

1. `Tools::SubagentPrompts#create_goal_with_pinned_task` synchronously creates a `user_message` Message via `Session#create_user_message(task)` so a `PinnedMessage` has something to attach to.
2. The new `child.enqueue_user_message(task)` queues a PendingMessage with the same content; drain promotes it as a **second** `user_message` Message.

The conversation history therefore has the task once near the top (the seed Message + its pin) and once after Melete enrichment (the promoted PM). Functional but smelly. The cleaner refactor is to drop the synchronous seed and pin lazily — e.g. via a subscriber that pins the first promoted `user_message` to the active root Goal. Out of scope for this PR; worth a follow-up ticket so the spawn path becomes single-entry.

## Adjacent tickets filed during smoke testing

- #451 — TUI tool_call line renders empty `$` for `bash` batch mode (`commands: [...]`, `mode: \"parallel\"`). Decorator branch handles `command` but not `commands`.
- #452 — TUI renders all PendingMessages with the user-message affordance. Now that PMs carry tool_response, subagent, Melete, and Mneme content (post-#439/#440), per-`message_type` rendering is needed.

## Test plan

- [x] Full suite: 2476 examples, 0 failures
- [x] `bundle exec standardrb` clean
- [x] New `PendingMessage` factory used across all touched specs
- [x] Smoke-tested parallel tool_use end-to-end — N=3 parallel Bash tool calls completed in a single drain cycle (one LLM call, all 3 tool_responses paired by tool_use_id)
- [x] Smoke-tested sub-agent self-start end-to-end — \`spawn_subagent\` → child runs full Mneme/Melete/drain/LLM → emits text → parent sees \`from_<nickname>\` tool pair → parent thinks. Lifecycle closes.

## Epic context

- Original brainstorm: \`thoughts/shared/notes/2026-04-03/qa-brainstorm-event-loop-mailbox.md\`
- Earned principles from this PR: \`thoughts/shared/notes/2026-04-19/event-driven-drain-principles.md\`
- Parallel tool_use AASM redesign: \`thoughts/shared/notes/2026-04-19/parallel-tool-use-aasm-redesign.md\`

---

> [!NOTE]
> **High Risk**
> High risk because it replaces the core LLM/tool execution loop and session state machine with a new event-driven pipeline, changes PendingMessage persistence/promotion semantics, and adds new jobs/events plus schema migrations that affect message ordering and recovery paths.
> 
> **Overview**
> **Replaces the imperative agent loop with an event-driven drain pipeline.** User input, tool results, sub-agent replies, and Mneme/Melete enrichment now enter via `PendingMessage`, are promoted in `DrainJob`, and a single LLM round-trip is processed via new events/subscribers (`StartMneme` → `StartMelete` → `StartProcessing` → `LLMResponded` → tool fan-out via `ToolExecutionJob`/`ToolExecuted`).
> 
> **Overhauls session claiming and tool-round closure.** `Session`’s AASM `start_processing` now also claims from `:executing` when `tool_round_complete?` is satisfied, uses locked transitions, and adds idle callbacks to clear interrupts and re-wake the drain when pending work exists.
> 
> **Removes legacy job/event infrastructure and updates integrations.** Deletes `AgentRequestJob`, `AgentLoop`, `MeleteJob`, and old tool-call/response events; updates spawn tools and sub-agent routing to enqueue pending messages; adds `PendingMessageDecorator`, new subscriber registrations, and DB migrations/indexes for `pending_messages` (`message_type` required, derived `kind`, plus `tool_use_id`, `success`, `bounce_back`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 00c36b29a9abb9af630fa0044b8a96512cbe4ce5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


app/jobs/drain_job.rb
```diff
@@ -0,0 +1,218 @@
...
+  discard_on Providers::Anthropic::AuthenticationError do |job, error|
+    session_id = job.arguments.first
+    Events::Bus.emit(Events::SystemMessage.new(
+      content: "Authentication failed: #{error.message}",
+      session_id: session_id
+    ))
+    ActionCable.server.broadcast(
+      "session_#{session_id}",
+      {"action" => "authentication_required", "message" => error.message}
+    )
```
hoblin: Should emit typed event as others. Broadcast should be triggered by subscriber . Atchitecture is the same as broadcast session state changes.

app/jobs/drain_job.rb
```diff
@@ -0,0 +1,218 @@
...
+  # @return [PendingMessage, nil] the active PM that was promoted (nil when
+  #   mailbox held only backgrounds or was entirely empty)
+  def promote_next(session)
+    actives = session.pending_messages.active
+    tool_response = actives.where(message_type: "tool_response").first
+    return promote_tool_response(tool_response) if tool_response
+
+    flush_backgrounds(session)
+    active = actives.order(:created_at).first
+    promote_active(active) if active
```
hoblin: Wrong order. Do not flush background messages if we don't have an active one that will run the LLM call.

app/jobs/drain_job.rb
```diff
@@ -0,0 +1,218 @@
...
+  end
+
+  def promote_tool_response(pm)
+    session = pm.session
```
hoblin: The job is a class that can have class variables. DRY the session, one job always work with only one session.

app/jobs/drain_job.rb
```diff
@@ -0,0 +1,218 @@
...
+        payload: {
+          "tool_name" => pm.source_name,
+          "tool_use_id" => tool_use_id,
+          "content" => content,
+          "success" => pm.success
+        },
+        timestamp: now_ns,
+        token_count: TokenEstimation.estimate_token_count(content)
+      )
+      pm.destroy!
```
hoblin: SOLID violation. This logic belongs to PM class.

app/jobs/drain_job.rb
```diff
@@ -0,0 +1,218 @@
...
+    end
+    pm
+  end
+
+  def flush_backgrounds(session)
+    session.pending_messages.background.find_each do |pm|
+      session.transaction do
+        session.promote_phantom_pair!(pm)
```
hoblin: Another SOLID violation. PM should promote itself. Move from the Session class.

app/jobs/drain_job.rb
```diff
@@ -0,0 +1,218 @@
...
+    session.transaction do
+      case pm.message_type
+      when "user_message"
+        session.create_user_message(pm.display_content, source_type: pm.source_type, source_name: pm.source_name)
+      when "subagent"
+        session.promote_phantom_pair!(pm)
+      else
+        raise "DrainJob cannot promote active PendingMessage #{pm.id} with message_type=#{pm.message_type.inspect}"
+      end
+      pm.destroy!
```
hoblin: Same here. PM.promote! should handle this. It knows it's type and should know what to do with it.

app/jobs/drain_job.rb
```diff
@@ -0,0 +1,218 @@
...
+    end
+    pm
+  end
+
+  def call_llm_and_emit(session)
+    client = build_client(session)
```
hoblin: def client
  @client ||= ...
end

You know the drill, we don't need a separate build method

app/jobs/drain_job.rb
```diff
@@ -0,0 +1,218 @@
...
+    registry = build_tool_registry(session)
```
hoblin: Same here. And with @session you don't need to pass it to every method. This is Ruby, not the fucking JS. Use OOP

app/jobs/drain_job.rb
```diff
@@ -0,0 +1,218 @@
...
+  end
+
+  # Bounce-back on the user's optimistically-committed message: delete
+  # the just-promoted Message so the TUI removes the phantom, emit
+  # {Events::BounceBack} so the client restores the text to the input.
+  # Non-bounce-back failures propagate to ActiveJob's retry machinery.
+  def bounce_back_or_reraise(session, promoted_pm, error)
+    raise error unless promoted_pm&.bounce_back?
+
+    last_msg = session.messages.where(message_type: "user_message").order(:id).last
```
hoblin: Dangerous. Could bounce wrong message or even several if job retries. Add attr_accessor to PM to store message id in memory. No need to persist it because PM is deleted afterwards, but it will help to match and bounce specific message only.

app/jobs/drain_job.rb
```diff
@@ -0,0 +1,218 @@
...
+      message_id: last_msg&.id
+    ))
+  end
+
+  # Ensures the session never sticks in +:awaiting+ — any path through
+  # {#perform} that exits while still awaiting an LLM response (success
+  # without tool_use lands in {Events::Subscribers::LLMResponseHandler},
+  # but our +ensure+ runs before that subscriber if we raised early)
+  # releases the claim so future kickoffs can pick up.
+  def release_from_awaiting(session)
```
hoblin: Already mentioned above, event listener should do this

app/jobs/drain_job.rb
```diff
@@ -0,0 +1,218 @@
...
+    else
+      LLM::Client.new
+    end
+  end
+
+  def build_tool_registry(session)
+    @shell_session = ShellSession.new(session_id: session.id)
+    restore_cwd(@shell_session, session)
+    Tools::Registry.build(session: session, shell_session: @shell_session)
+  end
```
hoblin: This belongs to Session class

app/jobs/drain_job.rb
```diff
@@ -0,0 +1,218 @@
...
+
+  def restore_cwd(shell_session, session)
+    cwd = session.initial_cwd
+    return unless cwd.present? && File.directory?(cwd)
+    shell_session.run("cd #{Shellwords.shellescape(cwd)}")
+  end
```
hoblin: Also does not belong to drain job. Make it a session method.

app/jobs/drain_job.rb
```diff
@@ -0,0 +1,218 @@
...
+    prompt = session.system_prompt
+    options[:system] = prompt if prompt
+    session.broadcast_debug_context(system: prompt, tools: registry.schemas)
+    options
+  end
+
+  def response_hash(response)
+    # Providers may wrap the response; unwrap so subscribers see a plain Hash.
+    response.respond_to?(:to_h) ? response.to_h.stringify_keys : response
+  end
```
hoblin: "May" might mean you did not check provider class and decided to add meaningless methos instead. ;)
Ensure it's needed.

app/jobs/drain_job.rb
```diff
@@ -0,0 +1,218 @@
...
+
+  def now_ns
```
hoblin: DRY violation. Check `ll ./config/initializers`

app/jobs/melete_enrichment_job.rb
```diff
@@ -0,0 +1,40 @@
...
+  queue_as :default
+
+  discard_on ActiveRecord::RecordNotFound
+
+  # @param session_id [Integer]
+  # @param pending_message_id [Integer, nil] the PM that kicked off the chain
+  def perform(session_id, pending_message_id: nil)
+    session = Session.find(session_id)
+
+    run_melete(session) if Anima::Settings.melete_blocking_on_user_message && !session.sub_agent?
```
hoblin: We didn't tackle this config setting. If I'm not mistaken setting it to false will break the call chain. Verify the flow if it's false and drop this setting if it breaks the new workflow. The initial idea was to make blocking enrichment optional. I'm not sure it makes sence now.

app/jobs/melete_enrichment_job.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  private
+
+  def run_melete(session)
+    Melete::Runner.new(session).call
+  rescue => error
+    msg = "FAILED (enrichment) session=#{session.id}: #{error.class}: #{error.message}"
+    Rails.logger.error("Melete #{msg}")
+    Melete.logger.error("#{msg}\n#{error.backtrace&.first(10)&.join("\n")}")
```
hoblin: Check what happens with the session and message if we hit this rescue. And ensure this is not the dead code we couldn't reach if runner have it's own exceptions handling.

app/jobs/mneme_enrichment_job.rb
```diff
@@ -0,0 +1,32 @@
...
+  # @param session_id [Integer]
+  # @param pending_message_id [Integer, nil] the PM that kicked off the chain
+  def perform(session_id, pending_message_id: nil)
+    session = Session.find(session_id)
+
+    begin
+      Mneme::PassiveRecall.new(session).call
+    rescue => error
+      Mneme.logger.error("PassiveRecall FAILED session=#{session_id}: #{error.class}: #{error.message}")
+    end
```
hoblin: Same question here. Is this rescue reachable and what happens with the session if we hit it. Follow the calls chain.

app/jobs/tool_execution_job.rb
```diff
@@ -0,0 +1,92 @@
...
+
+  # @param session_id [Integer]
+  # @param tool_use_id [String] Anthropic-assigned pairing ID
+  # @param tool_name [String]
+  # @param tool_input [Hash]
+  def perform(session_id, tool_use_id:, tool_name:, tool_input:)
+    session = Session.find(session_id)
+    @shell_session = ShellSession.new(session_id: session_id)
+    restore_cwd(@shell_session, session)
+    registry = Tools::Registry.build(session: session, shell_session: @shell_session)
```
hoblin: I'm sure i've seen this already. A good candidate for DRYing ;)

app/models/pending_message.rb
```diff
@@ -60,35 +59,54 @@ class PendingMessage < ApplicationRecord
...
-  # +user_message+ and +think+ enrich context first (Mneme → Melete →
-  # Processing); everything else bypasses enrichment and goes straight to
-  # the drain loop.
+  # User messages enrich context first (Mneme → Melete → Processing);
+  # tool responses and sub-agent deliveries bypass enrichment and go
+  # straight to the drain loop. Background message types route to nothing
+  # — they wait in the mailbox until an active turn drains them.
   MESSAGE_TYPE_ROUTES = {
     "user_message" => Events::StartMneme,
-    "think" => Events::StartMneme,
```
hoblin, line 74: The idea was that Ani's thinking (think tool call that produces response right away) would also go through enrichment, same way user message does. Do we still support this?

app/models/pending_message.rb
```diff
@@ -60,35 +59,54 @@ class PendingMessage < ApplicationRecord
...
   enum :kind, {background: "background", active: "active"}
 
+  before_validation :derive_kind
+
   validates :content, presence: true
-  validates :source_type, inclusion: {in: %w[user subagent skill workflow recall goal]}
+  validates :source_type, inclusion: {in: %w[user subagent skill workflow recall goal tool]}
   validates :source_name, presence: true, unless: :user?
-  validates :message_type, inclusion: {in: MESSAGE_TYPES}, allow_nil: true
+  validates :message_type, presence: true, inclusion: {in: MESSAGE_TYPES}
```
hoblin, line 105: Do we still need both source type and message type or we forgot to cleanup this?

app/models/pending_message.rb
```diff
@@ -129,6 +147,16 @@ def phantom_pair?
...
+  # Re-runs the {#route_to_event_bus} callback. Used by the idle-wake rule
+  # on {Session} to trigger the drain pipeline for messages that were
+  # created while the session was busy (and therefore saw a non-idle
+  # session in the original +after_create_commit+ and emitted nothing).
+  #
+  # @return [void]
+  def wake_drain_pipeline!
+    route_to_event_bus
+  end
+
```
hoblin: We don't need this alias especially with the bang. Rails have clear convention where to use `!` at the end of method names. This one just need better name that make sence everywhere we call it.

app/models/pending_message.rb
```diff
@@ -225,17 +253,30 @@ def broadcast_removed
...
+    event_class = MESSAGE_TYPE_ROUTES.fetch(message_type)
     Events::Bus.emit(event_class.new(session_id: session_id, pending_message_id: id))
   end
+
+  # Populates +kind+ from {MESSAGE_TYPE_KINDS} so callers only need to
+  # supply +message_type+. The mapping is the single source of truth for
+  # whether a message type triggers the drain loop or rides along as
+  # enrichment.
+  def derive_kind
+    return unless message_type
```
hoblin: Do we really have messages without a message_type? And if yes how enum with nil value will behave?

app/models/session.rb
```diff
@@ -636,17 +652,22 @@ def from_melete_messages
...
+  # {PendingMessage}. {DrainJob} flushes it into the conversation as a
+  # phantom tool_use/tool_result pair on the next drain cycle.
   #
   # @param source_type [String] "skill" or "workflow"
   # @param source_name [String] skill or workflow name
   # @param content [String] definition content to recall
   # @return [PendingMessage] the created pending message
   def enqueue_recall_message(source_type, source_name, content)
-    pending_messages.create!(content: content, source_type: source_type, source_name: source_name)
+    message_type = (source_type == "skill") ? "from_melete_skill" : "from_melete_workflow"
```
hoblin: I don't like this default. Switch case with default raise would be safer once we expand the system. Exception would inform us we missed the path.

lib/events/subscribers/tool_response_creator.rb
```diff
@@ -0,0 +1,37 @@
...
+      include Events::Subscriber
+
+      # @param event [Hash] Rails.event notification hash
+      def emit(event)
+        payload = event[:payload]
+        session_id = payload[:session_id]
+        return unless session_id
+
+        session = Session.find_by(id: session_id)
+        return unless session
```
hoblin: Another case of safety nets swallowing problems silently

lib/melete/runner.rb
```diff
@@ -167,10 +166,10 @@ def active_responsibilities
     #
     # @return [Array<Hash>] single-element messages array, or empty if no messages
     def build_messages
-      messages = recent_messages
-      return [] if messages.empty?
+      entries = recent_messages + pending_messages
+      return [] if entries.empty?
```
hoblin: Impossible case.

hoblin (commented):
Overall great job. Cudos Opus. Some minor issues need attention.

hoblin (commented):
## Re-Review Summary — PR #450

28 prior inline comments + Bugbot feedback in `44a5749`: the architectural asks (typed events, self-promoting PM, drop orphan tool_calls, AuthenticationBroadcaster, `promoted_message_id` tracking, `requires_lock`, `tool_round_complete?` guard) all landed correctly. The AASM collapse to a single `start_processing` event with guarded `:executing → :awaiting` is clean and matches the design notes.

A few real issues below — all introduced/persistent in the final commit, not regressions from review feedback.

### Major

**1. `AuthenticationError` is silently swallowed for bounce-back PMs** — `app/jobs/drain_job.rb:57-59`

```ruby
rescue => error
  release_after_failure(error) if @session
  raise error unless @active_pm&.bounce_back?
```

The in-perform rescue catches every error. For `user_message` PMs (`bounce_back?` → true) the error is swallowed, so `discard_on Providers::Anthropic::AuthenticationError` (line 39) never fires → `Events::AuthenticationRequired` is not emitted → TUI stays silent and the user can't re-enter the token. Same path hides any hard error behind a silent bounce on user messages.

Fix: either re-raise `AuthenticationError` unconditionally (let `discard_on` handle it), or emit `Events::AuthenticationRequired` from `Session#release_with_bounce_back` when the error is an auth error before consuming it.

**2. Missing indexes on new `pending_messages` columns** — `db/migrate/20260418150323_add_kind_and_message_type_to_pending_messages.rb` + `db/migrate/20260419120000_add_drain_fields_to_pending_messages.rb`

Both migrations add columns with zero indexes. `DrainJob#drain_mailbox` hits `pending_messages.where(message_type: \"tool_response\")` and `.where.not(message_type: \"tool_response\")` on every drain claim; `Session#tool_round_complete?` hits `pending_messages.where(message_type: \"tool_response\", tool_use_id: …)` under the AASM lock. Worth at least:

- composite `(session_id, message_type, created_at)` — matches `ordered_for_drain` + filter
- composite `(session_id, tool_use_id)` — matches the round-complete guard

**3. Drain pipeline subscribers/jobs have no unit specs** — no spec file for `DrainJob`, `ToolExecutionJob`, `MnemeEnrichmentJob`, `MeleteEnrichmentJob`, `LLMResponseHandler`, `ToolResponseCreator`, `DrainKickoff`, `AuthenticationBroadcaster`.

`Session` / `PendingMessage` specs cover the AASM + routing surface, and `SubagentMessageRouter` has its own spec. But the subscribers that own state transitions (LLMResponseHandler), fan-out (ToolExecutionJob), and error paths (DrainJob's transient retry + auth discard_on) are untested in isolation. The suite's green because the old AgentLoop specs were deleted alongside the code — the replacement isn't fully written yet. Unit coverage for the happy path + `may_*?` guarded paths + auth/transient error paths on each new piece would lock the contract.

### Minor

**4. `Session#tool_round_complete?` issues 3 queries per `:executing → :awaiting` transition** — runs inside `requires_lock: true`. Bounded by N parallel tools so not a live hot path, but collapsing to one SQL statement (or memoizing per `DrainJob#perform`) keeps the lock short.

**5. Stale doc in `lib/events/start_mneme.rb:4`** — YARD still says \"+user_message+ or +think+ PendingMessage\"; `think` is gone from `MESSAGE_TYPES`. Drop the reference.

### Nits

**6. `PendingMessage.ordered_for_drain` uses a raw `Arel.sql` clause** — safe today (enum-constrained, no interpolation) but `.order(Arel.sql(\"message_type = 'tool_response' DESC\")).order(:created_at)` is harder to break on the next refactor.

### Deliberate design — flagged by subagents, not issues

- `LLMResponseHandler#persist_agent_message` / `persist_tool_call` write `Message` records directly. This is the documented principle: PendingMessage is one-way inbound; LLM-produced content writes straight to Message and rides `MessageCreated` to the lifecycle subscribers (`SubagentMessageRouter` picks it up via `MessageCreated` filter on `message_type == \"agent_message\"`). Not a three-layer violation.
- `with_transient_retry` blocking the worker thread during backoff is an accepted trade-off — inline retry is required because `PendingMessage#promote!` destroys rows before the LLM call, so a job-level `retry_on` would no-op. Docstring at `drain_job.rb:34-37` covers this.
- Sub-agent task double-seeding (sync Message for the pin + PM promotion) is called out in the PR body as a known smell with a follow-up. Out of scope.

---

*Review generated with multi-agent analysis + judgment pass against the earned-principles notes.*

bonk-moltbot (approved):
🐭✅ Magnificent.

This is a full architecture replacement — imperative `AgentLoop` → event-driven pipeline — and it reads like it was designed in one sitting. The mermaid diagrams in the PR body are worth their weight in gold.

**What makes this work:**

- **AASM as the sole arbiter.** `start_processing` collapsing idle-claim and round-close into one atomic locked transition is the key insight. No race conditions, no double-claims. `tool_round_complete?` as a guard means the AASM knows when the round is whole — not the drain, not the tool executor.

- **Event emission = final act.** Every piece does its work, emits, and exits. DrainJob never transitions past `:awaiting`. LLMResponseHandler never calls the LLM. ToolExecutionJob never creates PendingMessages. Clean separation.

- **PendingMessage as single inbound path.** Users, tools, sub-agents, Mneme, Melete — all enter through the same mailbox. `MESSAGE_TYPE_KINDS` as the single source of truth for active-vs-background. `derive_kind` from `before_validation` — callers never think about kind.

- **Tool responses bulk-promote.** The old per-PM promotion that silently corrupted history under parallel tool_use is gone. The drain picks up ALL tool_responses of the completed round in one transaction.

- **Sub-agent self-start through the same pipeline as humans.** `spawn_subagent` enqueues a user_message PM on the child → Mneme → Melete → Drain. No special-case spawn-side drain triggering. The reply path via `SubagentMessageRouter` → `parent.enqueue_user_message` closes the loop with the same vocabulary.

- **ToolExecutionJob rescue → always emits.** A crash can never leave the session stuck in `:executing`. Synthetic failure event ensures ToolResponseCreator runs.

- **`wake_drain_pipeline_if_pending` on idle transition.** The race where a PM queues while the session is `:awaiting` is covered. Clean.

**The known smell is honest.** Double-seeded sub-agent task (synchronous Message + PM) — documented, understood, follow-up worthy. Not blocking.

**Dead code removal is thorough:** AgentLoop, AgentRequestJob, MeleteJob, cargo-cult ToolCall/ToolResponse events, `between_rounds`, `chat_with_tools` coupling — all gone. Net AASM events: 5 → 4.

2476 examples, 0 failures, standardrb clean. Smoke-tested parallel tool_use and sub-agent lifecycle end-to-end.

Issue #453, opened 2026-04-19
Shell session: spawn login shell for full user environment


bonk-moltbot:
## Problem to solve

The agent's bash shell is spawned with `--norc --noprofile`, which strips all user environment: PATH, custom env vars, tool locations. The agent can't reach tools documented as "in the PATH" (#434), and the shell environment doesn't match what the human sees in their terminal.

The root cause is one line in `ShellSession#spawn_shell`:

```ruby
PTY.spawn(SHELL_ENV, "bash", "--norc", "--noprofile")
```

The agent's shell should be identical to the user's shell. Same PATH, same env, same tools. Like VSCode's integrated terminal — it spawns a login shell and it just works.

## Solution

Spawn the user's actual login shell instead of a stripped-down bash.

### Implementation

Replace the current spawn:

```ruby
# Before
PTY.spawn(SHELL_ENV, "bash", "--norc", "--noprofile")

# After
shell = ENV["SHELL"] || "/bin/bash"
PTY.spawn(SHELL_ENV, shell, "-l")
```

That's it. The `SHELL_ENV` overrides (`TERM=dumb`, `PAGER=cat`, `GIT_PAGER=cat`, etc.) still merge on top, preventing pagers and interactive prompts. But everything from the user's shell profile loads normally — PATH, aliases, env vars, tool locations.

The `init_shell` method already handles any startup noise: it sets `PS1=''`, redirects stderr to the FIFO, and `consume_until(marker)` discards everything before the init marker. Welcome messages, motd, shell startup output — all consumed and discarded.

### What about marker parsing?

The marker-based output parsing (`__ANIMA_<hex>__`) works on stdout. Shell profiles write to stderr (warnings) or set env vars (silent). Even if something writes to stdout during init, `consume_until` in `init_shell` eats it all before the first real command runs. No breakage.

### What about zsh vs bash?

If the user's `$SHELL` is zsh, we spawn zsh. The command syntax we send (`cmd; echo marker $?`) is POSIX-compatible — works in bash, zsh, fish, dash, anything. The agent gets the exact same environment the user has.

## Scope

- Modify `ShellSession#spawn_shell`: use `$SHELL -l` instead of `bash --norc --noprofile`
- Keep `SHELL_ENV` overrides (TERM, PAGER, etc.) — these are safety guards, not environment restrictions
- Verify #433 is resolved (cwd persists across calls — already works via PTY)
- Verify #434 is resolved (user PATH available)

## Out of scope

[...]

PR #456, opened 2026-04-19
#453 feat: persistent shell session + Mneme sanitization fix

### LINKED ISSUE #433: Bash tool reports working directory persistence but doesn't actually persist it
## Problem to solve

The `bash` tool's description states: *"Working directory and environment persist between calls."* After a `cd`, the tool even announces `You are now in /some/path` in its output. But the next `bash` call starts back in the original cwd — each invocation is an independent shell.

This is actively misleading: the announcement makes the agent confident state persisted when it didn't. As a workaround I've been prefixing every command with `cd ~/Work/anima && …`, which bloats history, makes `| tail` and exit-code handling fragile, and broke my ability to trust the tool's own status line.

## Reproduction

```
bash: cd ~/Work/anima
→ "You are now in /home/hoblin/Work/anima, git repo hoblin/anima on branch …"

bash: pwd
→ /home/hoblin/.local/share/mise/installs/ruby/3.4.8/lib/ruby/gems/3.4.0/gems/anima-core-1.4.0
```

The second call is back in the gem install directory (which is where the shell starts by default). No persistence happened.

## Solution

Two acceptable fixes — either is fine, they're mutually exclusive:

1. **Make it honest.** Actually persist cwd (and env) between calls — wrap each call in a long-lived shell process or track cwd server-side and `cd` into it before each command.
2. **Fix the docs and the status line.** Remove "Working directory and environment persist between calls" from the description, stop emitting `You are now in …` after `cd`, and document that each call is an independent shell so agents know to prefix.

Option 1 matches the current description and is what agents naturally assume. Option 2 is simpler but requires every agent to learn the workaround.

## Why this matters

This isn't cosmetic. When I was auditing `next_view_mode` earlier today, I ran `bundle exec rspec spec/models/session_spec.rb` (prefixed with `cd`, piped to `tail`) and got 19 failures. I reported them as pre-existing. Eugene ran the same command without the prefix and got a clean run — different Ruby/bundler environment because his shell was actually in `~/Work/anima` with mise shims loaded, and mine wasn't. I called specs "broken" based on a state that didn't reflect reality.

The bug cost us trust in a test result. Worth fixing.

## Environment

- Anima v1.4.0
- Bash tool as shipped in current Claude Code / Anima agent

---
🧬 [Ani](https://github.com/hoblin/anima) (anima-core v1.4.0)

### LINKED ISSUE #434: Agent's bash shell inherits a minimal PATH from foreman, missing tools documented as available
## Problem to solve

The agent's `bash` tool runs commands in a shell spawned from a Ruby process started by `foreman` under systemd. That shell inherits systemd's minimal environment — no `~/.zshrc`, `~/.zprofile`, or any interactive shell init is sourced. As a result, the agent's `PATH` is missing entries the human's interactive terminal has, and tools documented as "in the PATH" by built-in workflows are unreachable.

Today the `create_handoff` workflow told me to run `spec-metadata` "in the PATH". It exists at `/home/hoblin/thoughts/bin/spec-metadata` and the human can run it from their tmux terminal. I can't — my shell has no `/home/hoblin/thoughts/bin`, no `/home/hoblin/.local/bin`, no `./bin`. I fell back to gathering the metadata manually.

## Reproduction

```
bash tool: echo $PATH | tr ':' '\n'
→ mise shims + system paths only, no ~/thoughts/bin, no ~/.local/bin, no ./bin

bash tool: pstree -ps $$
→ systemd(1)→systemd(user,1539)→foreman→ruby→bash
```

Compare to the human's interactive zsh in tmux:

```
~/thoughts/bin                ← spec-metadata lives here
~/.local/bin
./bin                         ← per-repo bin override
... everything the agent has, plus oh-my-zsh plugin dirs
```

## Root cause

`foreman` inherits PATH from the systemd user service that launches it. Anima's brain process (and therefore the agent's bash tool) never sees `~/.zshrc`. The `Procfile` / systemd unit doesn't export a curated PATH, so the agent ends up with whatever `systemd --user` gives it by default.

## Solution options

1. **Curate an explicit PATH in the systemd unit or `Procfile`** — set `Environment=PATH=...` to include `~/.local/bin`, `~/thoughts/bin`, `~/bin`, and the mise shims. Explicit, reviewable, no surprises.
2. **Wrap commands in a login shell** — e.g. `zsh -lc '...'` or `bash -lc '...'`. Works but is slow (sources rc files on every call) and couples the brain to the human's shell choice.
3. **Ship an `anima-agent.env` file** the brain reads at boot and merges into the agent's shell env. Cleanest long-term: the human declares what the agent should have, versioned with the project.

I'd pick **(1) for the quick fix and (3) for the long-term** — declare the agent's environment explicitly rather than inheriting it.

## Why this matters

Two bugs surfaced from this same root cause today:

- The `create_handoff` workflow references `spec-metadata`, a tool that's in the human's PATH but not the agent's. The workflow docs are wrong for the agent, and I had to fall back to manual metadata gathering.
- Possibly the rspec discrepancy in #433 — I was running `bundle exec rspec` in a shell whose PATH might not resolve bundler/mise the same way the human's does. Worth verifying once this is fixed.

The broader principle: **the agent's environment should be declared, not inherited.** Right now what works for the agent depends on what systemd happened to pass to foreman, which is fragile and non-obvious.

## Environment

- Anima v1.4.0
- Process tree: `systemd(1) → systemd(user, 1539) → foreman → ruby → bash`
- Human shell: zsh with antigen + oh-my-zsh
- Agent shell PATH: mise shims + `/usr/local/*`, `/usr/*`, nothing else

---
🧬 [Ani](https://github.com/hoblin/anima) (anima-core v1.4.0)

### LINKED ISSUE #457: Mneme FTS5 crashes drain pipeline on hyphenated user input
## Problem to solve

Whenever a user message (or subagent task prompt) contains a hyphenated word — `sub-agents`, `event-driven`, `out-of-scope`, anything — the `MnemeEnrichmentJob` raises `SQLite3::SQLException: no such column: agents`, which halts the drain pipeline for that session. `StartMelete` never emits, `StartProcessing` never emits, `DrainJob` never runs, the user's message sits stranded in the mailbox forever. Subagents are hit the hardest because their *first and only* message often contains such wording — so they spawn, go `idle`, and never respond. The pipeline is silently load-bearing on "no one ever types a hyphen."

## Root cause

Two independent bugs compound:

1. `Mneme::Search#sanitize_query` kept hyphens inside tokens (`[^a-zA-Z0-9-]`). FTS5 parses a hyphen between identifiers as the `NOT` operator, so `sub-agents` turns into `sub NOT agents`. The `messages_fts` virtual table has one column (`searchable_text`), so the bare `agents` was then resolved as a **column name** → `no such column: agents`.
2. `MnemeEnrichmentJob` re-raised the error. Mneme recall is *enrichment* — recalled memories are a nice-to-have — but the job's failure blocked the handoff to `StartMelete`, taking the primary pipeline down with it.

## Reproduction

```ruby
session = Session.find(...)
session.enqueue_user_message("tell me about sub-agents")
# → MnemeEnrichmentJob raises, session stays idle, message stranded
```

Or: ask Ani to spawn a subagent whose task includes the word `sub-agents`. The subagent spawns, hits `idle`, never processes.

## Solution

Two-part fix:

- **Make Mneme non-fatal for the pipeline.** `MnemeEnrichmentJob` now rescues and logs recall failures, then still emits `StartMelete` so the user message reaches the LLM even if recall blew up.
- **Sanitize FTS5 queries the canonical way.** Follow SQLite authors' recommendation: quote-wrap each token (inside `""` FTS5 does not parse any operators), with a small allowlist of logical operators (`AND`, `OR`, `NOT`, `NEAR`) passed through unchanged so callers that build `word1 OR word2` queries still get boolean behavior.

## Why quote-wrapping, not a blocklist?

The first fix attempt was to strip unsafe characters (`-`, `:`, `*`, `^`, `{`, `}`, `(`, `)`). Research into FTS5 best practices flagged this as incomplete: a blocklist has to be re-audited against every FTS5 release, and even stripping `-` still leaves `agents:foo` injecting a column filter. Quote-wrapping is the approach SQLite's own authors recommend ([mailing list](https://www.mail-archive.com/sqlite-users@mailinglists.sqlite.org/msg118320.html), [FTS5 docs](https://www.sqlite.org/fts5.html)) — it's structural, future-proof, and one-line to extend if a caller legitimately needs a new operator.

## Fixed in

[#456](https://github.com/hoblin/anima/pull/456)

hoblin:
## Summary

Delivers real per-conversation shell persistence, inherits the user's login environment — what #433 and #434 actually asked for — and fixes a latent FTS5 crash that stranded subagents mid-spawn (#457). The issue's original pseudo-code (\`\$SHELL -l\`) didn't account for the PTY/interactive-shell collision or the fact that the existing job wiring tore the shell down after every \`perform\`, so this PR takes a wider cut:

### Shell persistence (#433, #434, #453)

- **Login environment inherited once per conversation.** \`ShellSession\` spawns \`\$SHELL -l -c 'exec bash --norc --noprofile'\`. The outer login shell sources the user's profile (PATH, mise, tool locations); \`exec\` hands off to a bare bash so the PTY sees clean, predictable I/O — no ZLE, no starship, no bracketed paste, no ANSI escapes corrupting our marker parser. Falls back to \`/bin/bash\` when \`\$SHELL\` is unset/empty.
- **Session-keyed shell registry.** \`ShellSession.for_session(session)\` now caches one live shell per \`Session#id\` for the process's lifetime. First call spawns + \`cd\`s into \`initial_cwd\`; every subsequent call reuses the same bash. Dead cached shells are replaced transparently. \`release(session_id)\` is available for explicit eviction; \`at_exit\` still sweeps everything on shutdown.
- **Jobs no longer finalize per-\`perform\`.** \`ToolExecutionJob\` and \`DrainJob\` drop the \`ensure { @shell_session&.finalize }\` line — the shell outlives the job, which is the whole point.

### Mneme FTS5 hardening (#457) — surfaced while smoke-testing subagents

- **Mneme recall is now non-fatal for the pipeline.** \`MnemeEnrichmentJob\` rescues + logs recall failures, then still emits \`StartMelete\`. Recall is enrichment; its failure must not gate the primary drain. Previously, a Mneme crash left the session \`idle\` with its user_message stranded forever.
- **FTS5 query sanitization rewritten per SQLite guidance.** \`Mneme::Search#sanitize_query\` now quote-wraps each token — inside \`\"\"\` FTS5 treats every operator character as literal content — with a small allowlist of logical operators (\`AND\`, \`OR\`, \`NOT\`, \`NEAR\`) passed through so callers that build \`word1 OR word2\` queries still get boolean behavior. This replaces an incomplete char-level blocklist. The triggering bug: a user message containing \`sub-agents\` made FTS5 parse \`sub NOT agents\`, then try to resolve \`agents\` as a column on the single-column \`messages_fts\` table → \`no such column: agents\` → whole pipeline halts.

### Constraints worth naming

- **Worker affinity.** The ShellSession registry is in-process. Today the brain is a single-process setup (\`bin/dev\`: one Puma + embedded SolidQueue), so this is fine. If we ever split into multiple worker processes, tool calls for the same session would need session-affinity routing or shared IPC to keep state coherent. Out of scope here.
- **Login shell choice of option 2 from #434.** The author of #434 flagged \"slow — sources rc files on every call\" as the downside; with real persistence that cost is amortized to once per conversation, so the critique no longer applies. Option 1 (curate PATH explicitly in systemd) and option 3 (\`anima-agent.env\`) remain reasonable future tightenings if we want the agent's environment *declared* instead of *inherited*.

## Test plan

- [x] \`bundle exec rspec spec/lib/shell_session_spec.rb\` — 59 pass, including specs for \`.for_session\` cache hit / miss / dead-shell replacement / session isolation and \`.release\`
- [x] \`bundle exec rspec spec/lib/mneme/ spec/jobs/ spec/lib/tools/search_messages_spec.rb\` — 204 pass, including new regressions for hyphen-as-NOT and colon-as-column-filter
- [x] \`bundle exec standardrb --fix\` — clean
- [x] \`bundle exec reek lib/shell_session.rb lib/mneme/search.rb\` — no new smells introduced
- [x] Live smoke test with Ani on the dev brain (port 42135):
  - \`cd /tmp && pwd\` → \`/tmp\`, then separate \`pwd\` call → \`/tmp\` ✅
  - \`export ANIMA_PERSIST_TEST='hello from call 1'\`, then separate \`echo \$ANIMA_PERSIST_TEST\` → \`hello from call 1\` ✅
- [ ] Live smoke test: subagent spawn with hyphenated task prompt now completes cleanly
- [ ] CI matrix green

Closes #433
Closes #434
Closes #457
Refs #453

---

> [!NOTE]
> **Medium Risk**
> Changes shell lifecycle to be persistent and session-scoped and adjusts Mneme recall/search behavior; mistakes could leak processes, cross-contaminate sessions, or alter recall results and pipeline progress.
> 
> **Overview**
> **Adds true per-conversation shell persistence.** `ShellSession.for_session` now caches one live shell per `Session#id`, spawns via the user’s login shell (`$SHELL -l`) then `exec`s into a bare `bash`, and introduces `.release` plus safer registry locking/teardown to manage lifetimes and crashes.
> 
> **Jobs stop tearing down shells per run.** `DrainJob` and `ToolExecutionJob` no longer `finalize` shells in `ensure`, so working directory and exported env vars survive across tool calls within a session.
> 
> **Hardens Mneme so recall can’t strand the drain and FTS5 input can’t crash.** `MnemeEnrichmentJob` now rescues/logs recall failures but still emits `StartMelete`, and `Mneme::Search` rewrites query sanitization to quote-wrap tokens with an allowlist of boolean operators, preventing hyphen/colon/quote-induced FTS5 syntax errors; specs updated/added accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 36c0036a02ddac750d3ed6a0f3c381ca012a0353. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


bonk-moltbot (approved):
🐭✅

Three fixes, one PR, zero waste.

**Shell persistence:** `$SHELL -l -c 'exec bash --norc --noprofile'` is the cleanest way to get user env into a predictable PTY. Login shell sources profiles → exec hands off → no ZLE/starship/bracketed-paste noise. `Monitor` over `Mutex` for reentrant locking is the right call — `for_session` → `new` → `register` is a textbook reentry path.

**FTS5 sanitization:** Quote-wrapping tokens is exactly what SQLite docs recommend. The old char-level blocklist would need auditing on every FTS5 release; this is future-proof. Good catch on the `sub-agents` → `sub NOT agents` edge case.

**Mneme non-fatal:** Recall is enrichment, not gating. Logging loudly + continuing the pipeline is the correct trade-off. A stranded user_message with no retry trigger was a real production risk.

PR #461, opened 2026-04-20
#452 feat: per-message_type rendering for PendingMessages

### LINKED ISSUE #452: TUI: per-kind rendering for PendingMessage
## Problem to solve

PendingMessage was originally designed to carry only user messages, so the TUI renders all pending messages (dimmed) with user-message affordances — a single visual treatment regardless of content type.

Since #439 and #440, PendingMessage now also carries `tool_response`, sub-agent responses, and `from_melete_*`/`from_mneme` background messages. Rendering all of these as dimmed user messages is misleading — a tool response looks identical to a user prompt, and Melete skill injections are indistinguishable from something the user typed.

## Solution

Introduce per-`message_type` rendering for pending messages in the TUI. Each PendingMessage kind/type should get the same visual treatment its promoted counterpart would receive, just dimmed or marked as pending.

## Architecture Context

The existing decorator architecture already supports per-type rendering — this ticket extends it to PendingMessages.

### Server side (Rails/Draper) — per message TYPE

- `Message#decorator_class` (`app/models/message.rb:99-111`) overrides Draper's default resolution to pick the correct subclass decorator by `message_type`
- Each concrete subclass (`ToolCallDecorator`, `UserMessageDecorator`, etc.) handles one message type
- `PendingMessageDecorator` (`app/decorators/pending_message_decorator.rb`, PR #450) exists but currently treats all PMs uniformly
- View mode (`basic/verbose/debug`) on Session controls what data enters the ActionCable wire — decorators enforce this boundary

### TUI client side — per TOOL NAME

- `build_rendered_lines` in `lib/tui/screens/chat.rb` dispatches rendering by `role` via `when` clauses
- Each role maps to a `render_*_entry` method
- **Critical gotcha**: Every new role value needs a matching `when` clause + `render_*_entry` method, or it silently falls through to the `else` clause and renders as plain white text — no error, no crash

### The split

| Side | Decorator type | Granularity | Job |
|------|---------------|-------------|-----|
| **Server (Rails/Draper)** | Per message TYPE | Uniform across all tools of same type | Controls WHAT data enters the ActionCable wire |
| **TUI client** | Per TOOL NAME | Unique per tool | Controls HOW it looks on screen (colors, layout, detail) |

### 1. Server side: `PendingMessageDecorator` per-type dispatch

`PendingMessageDecorator` should key off `message_type` to produce type-appropriate JSON payloads, mirroring how `Message#decorator_class` picks the right subclass. Options:

- **Option A**: `PendingMessage#decorator_class` override (same pattern as `Message`) routing to type-specific PM decorators
- **Option B**: Single `PendingMessageDecorator` with internal dispatch by `message_type` (simpler if PM payloads stay lightweight)

### 2. TUI side: `when` clauses for PM message types

`build_rendered_lines` needs matching `when` clauses for each PM type:

- **user message** — current dimmed treatment (no change)
- **tool_response** — tool result styling (bordered block, code formatting), dimmed
- **sub-agent response** (`from_sleuth`, `from_scout`, etc.) — sub-agent badge/prefix, dimmed
- **from_melete_skill / from_melete_workflow / from_melete_goal** — Melete badge/prefix, dimmed
- **from_mneme** — Mneme badge/prefix, dimmed

All PM types render dimmed to distinguish from promoted Messages.

## Key files

- `app/models/message.rb:99-111` — `Message#decorator_class` (pattern to follow)
- `app/decorators/message_decorator.rb` — base decorator with dispatcher + helpers
- `app/decorators/pending_message_decorator.rb` — current uniform PM decorator
- `app/models/concerns/message/broadcasting.rb` — AR callbacks triggering broadcasts
- `app/channels/session_channel.rb:169-176` — `decorate_event_payload` (where decorators hit the wire)
- `lib/tui/screens/chat.rb` — `build_rendered_lines` with `when` dispatch by role

## Reference docs

- `thoughts/shared/notes/2026-03-06/draper-as-universal-event-representation.md` — original vision
- `thoughts/shared/notes/2026-03-09/tui-view-modes-draper-decorators.md` — TUI design spec
- `thoughts/shared/notes/2026-03-13/draper-decorators-in-event-driven-client-server-architecture.md` — canonical architecture note (flagged as load-bearing in handoffs)

## Why per-kind rendering instead of just hiding non-user PMs?

Visibility matters for debugging and for understanding what's entering the context window. The drain-loop architecture (#427) routes *everything* through PendingMessage — hiding non-user types would obscure the pipeline. Dimmed type-specific rendering preserves transparency while keeping the visual hierarchy clear.

Part of #427

hoblin:
Closes #452. Part of epic #427.

## Problem

Post #439/#440 the PendingMessage mailbox carries five distinct content kinds — user input, tool responses, sub-agent replies, Mneme recalls, Melete activations (skill / workflow / goal). The TUI rendered every one of them with the user-message affordance (dimmed), so a tool result was indistinguishable from something the user typed and Melete's skill injections looked like inbound prompts. The drain pipeline became opaque just as it grew the most surface area worth showing.

## Solution

Push per-`message_type` rendering through the same Draper architecture promoted Messages already use.

### Server: subclass-per-type decorators

`PendingMessageDecorator` becomes an abstract base mirroring `MessageDecorator` — `RENDER_DISPATCH` over basic/verbose/debug/melete/mneme, abstract `render_basic`, default delegations.

Concrete subclasses:

| message_type | decorator | basic | verbose+ |
|---|---|---|---|
| `user_message` | `PendingUserMessageDecorator` | dimmed user line | + delegates to basic |
| `tool_response` | `PendingToolResponseDecorator` | nil (matches `ToolResponseDecorator`) | dimmed tool_response payload, 3-line cap |
| `subagent` | `PendingSubagentDecorator` | nil | `[from <nickname>]` badge, dimmed |
| `from_mneme` | `PendingFromMnemeDecorator` | nil | `[Mneme recall]` badge, dimmed |
| `from_melete_skill` | `PendingFromMeleteSkillDecorator` | nil | `[Melete skill: <source>]` badge |
| `from_melete_workflow` | `PendingFromMeleteWorkflowDecorator` | nil | `[Melete workflow: <source>]` badge |
| `from_melete_goal` | `PendingFromMeleteGoalDecorator` | nil | `[Melete goal: <source>]` badge |

`PendingMessage#decorator_class` mirrors `Message#decorator_class` — `pm.decorate` resolves to the right subclass automatically. Each payload carries `status: \"pending\"` so the TUI dims it. Existing Mneme/Melete transcript lines are preserved (each subclass owns its `render_melete` / `render_mneme`).

### Wire: decorated payload on broadcast

`PendingMessage#broadcast_payload(mode)` emits the structured payload exactly like `MessageBroadcaster` does for promoted messages. `broadcast_created`, `transmit_history`, and `broadcast_viewport` all flow through it — the TUI receives the rendered hash now, not just bare `content`.

### TUI: dispatch + dimming

- `MessageStore#add_pending` accepts the full payload, extracts `rendered`, and **drops the entry when a decorator hides the PM in the current mode** (so a verbose→basic switch strips non-conversational pendings cleanly).
- `chat.rb#build_rendered_lines` adds `when` clauses for the three new pending roles (`pending_subagent`, `pending_melete`, `pending_mneme`) routing to a shared `render_pending_badge_entry` helper.
- `lib/tui/decorators/base_decorator.rb` honors `status: \"pending\"` via `effective_color` / `effective_response_color`, so per-tool subclass colors stay intact when active and dim uniformly when pending — without each subclass having to know about the pending state.

## Test plan

- [x] Full suite: 2450 examples, 0 failures
- [x] \`bundle exec standardrb\` clean
- [x] Per-decorator specs for every subclass cover render_basic/verbose/debug shape + dispatch resolution
- [x] MessageStore + chat screen specs cover the structured payload form, the hide-on-mode-change path, and per-type dispatch
- [ ] Live TUI smoke test (needs an LLM-driven session to surface non-user PMs end-to-end) — follow-up

---

> [!NOTE]
> **Medium Risk**
> Changes the ActionCable wire payload and both server/client rendering paths for `PendingMessage`s, so regressions could hide or misrender in-flight messages across view modes. Logic is UI/presentation-focused but touches real-time session streaming and decorator dispatch.
> 
> **Overview**
> Pending mailbox items are now rendered **per `PendingMessage#message_type`** instead of always looking like dimmed user input.
> 
> On the server, `PendingMessageDecorator` becomes an abstract base with new concrete pending decorators (user, tool_response, subagent, Mneme recall, and Melete skill/workflow/goal) and `PendingMessage#decorator_class` picks the right one; pending broadcasts/transmit history now send a structured `rendered` payload via `PendingMessage#broadcast_payload` (including `message_type`) rather than raw `content`.
> 
> On the TUI side, `MessageStore#add_pending` consumes the full payload, drops entries when the decorator hides them for the current mode, and normalizes keys; `chat.rb` adds rendering for `pending_subagent`/`pending_mneme`/`pending_melete` badges, and tool decorators dim call/response colors when `status: "pending"` is present. Tests are updated/added to cover decorator dispatch, payload shape, mode-hiding behavior, and pending badge rendering, and the Mneme recall prompt copy is tightened.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 528eb7dde53c584ba28717f93e2d82c488f9b209. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/README.md
+++ b/README.md
@@ -444,8 +444,8 @@ Three switchable view modes let you control how much detail the TUI shows. Cycle
 View modes are implemented as a three-layer decorator architecture:
 
 - **ToolDecorator** (server-side, pre-event) — transforms raw tool responses for LLM consumption. Content-Type dispatch converts HTML → Markdown, JSON → TOON. Sits between tool execution and the event stream.
-- **EventDecorator** (server-side, Draper) — uniform per event type (`UserMessageDecorator`, `ToolCallDecorator`, etc.). Decides WHAT structured data enters the wire for each view mode.
-- **TUI Decorator** (client-side) — unique per tool name (`BashDecorator`, `ReadDecorator`, `EditDecorator`, etc.). Decides HOW each tool looks on screen — tool-specific icons, colors, and formatting.
+- **EventDecorator** (server-side, Draper) — uniform per message type (`UserMessageDecorator`, `ToolCallDecorator`, etc.) for promoted messages, and a parallel family (`PendingUserMessageDecorator`, `PendingToolResponseDecorator`, `PendingSubagentDecorator`, `PendingFromMnemeDecorator`, `PendingFromMelete{Skill,Workflow,Goal}Decorator`) for in-flight `PendingMessage` rows. Decides WHAT structured data enters the wire for each view mode; pending payloads carry `status: "pending"` so the TUI dims them.
+- **TUI Decorator** (client-side) — unique per tool name (`BashDecorator`, `ReadDecorator`, `EditDecorator`, etc.). Decides HOW each tool looks on screen — tool-specific icons, colors, and formatting. Honors `status: "pending"` to render in-flight content in the muted theme color.
 
 Mode is stored on the `Session` model server-side, so it persists across reconnections.
 
--- a/app/channels/session_channel.rb
+++ b/app/channels/session_channel.rb
@@ -262,7 +262,7 @@ class SessionChannel < ApplicationCable::Channel
     end
 
     session.pending_messages.find_each do |pm|
-      transmit({"action" => "pending_message_created", "pending_message_id" => pm.id, "content" => pm.content})
+      transmit(pm.broadcast_payload(session.view_mode))
     end
   end
 
@@ -281,7 +281,7 @@ class SessionChannel < ApplicationCable::Channel
     end
 
     session.pending_messages.find_each do |pm|
-      ActionCable.server.broadcast(stream_name, {"action" => "pending_message_created", "pending_message_id" => pm.id, "content" => pm.content})
+      ActionCable.server.broadcast(stream_name, pm.broadcast_payload(session.view_mode))
     end
   end
 
--- a/app/decorators/pending_from_melete_decorator.rb
+++ b/app/decorators/pending_from_melete_decorator.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+# Shared base for the three Melete-activation pending decorators (skill,
+# workflow, goal). All three share the same TUI shape — a dimmed
+# +pending_melete+ payload with +kind+ + +source+ + truncated content
+# — and only differ on the +KIND+ constant and the per-type Melete
+# transcript line. Subclasses override +KIND+ and +render_melete+; this
+# base owns everything else.
+class PendingFromMeleteDecorator < PendingMessageDecorator
+  # @return [nil] Melete activations are hidden in basic mode
+  def render_basic
+    nil
+  end
+
+  # @return [Hash] dimmed Melete-activation payload
+  def render_verbose
+    {
+      role: :pending_melete,
+      kind: self.class::KIND,
+      source: source_name,
+      content: truncate_lines(content, max_lines: 3),
+      status: "pending"
+    }
+  end
+
+  # @return [Hash] full Melete-activation payload
+  def render_debug
+    {
+      role: :pending_melete,
+      kind: self.class::KIND,
+      source: source_name,
+      content: content,
+      status: "pending"
+    }
+  end
+end
--- a/app/decorators/pending_from_melete_goal_decorator.rb
+++ b/app/decorators/pending_from_melete_goal_decorator.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+# Decorates a +from_melete_goal+ {PendingMessage} — a goal event Melete
+# logged for the upcoming turn (created/updated/closed). See
+# {PendingFromMeleteSkillDecorator} for the parallel design.
+class PendingFromMeleteGoalDecorator < PendingMessageDecorator
+  KIND = "goal"
+
+  # @return [nil] Melete activations are hidden in basic mode
+  def render_basic
+    nil
+  end
+
+  # @return [Hash] dimmed Melete-goal activation payload
+  def render_verbose
+    {
+      role: :pending_melete,
+      kind: KIND,
+      source: source_name,
+      content: truncate_lines(content, max_lines: 3),
+      status: "pending"
+    }
+  end
+
+  # @return [Hash] full Melete-goal activation payload
+  def render_debug
+    {
+      role: :pending_melete,
+      kind: KIND,
+      source: source_name,
+      content: content,
+      status: "pending"
+    }
+  end
+
+  # @return [String] Melete transcript line — goal id and content
+  def render_melete
+    "Melete logged goal #{source_name}: #{truncate_middle(content)}"
+  end
+end
--- a/app/decorators/pending_from_melete_goal_decorator.rb
+++ b/app/decorators/pending_from_melete_goal_decorator.rb
@@ -2,36 +2,9 @@
 
 # Decorates a +from_melete_goal+ {PendingMessage} — a goal event Melete
 # logged for the upcoming turn (created/updated/closed). See
-# {PendingFromMeleteSkillDecorator} for the parallel design.
-class PendingFromMeleteGoalDecorator < PendingMessageDecorator
+# {PendingFromMeleteDecorator} for the shared TUI rendering shape.
+class PendingFromMeleteGoalDecorator < PendingFromMeleteDecorator
   KIND = "goal"
-
-  # @return [nil] Melete activations are hidden in basic mode
-  def render_basic
-    nil
-  end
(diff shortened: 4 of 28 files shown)


hoblin (commented):
## PR Review Summary

**Verdict: APPROVE** ✅

The architecture is clean — mirroring `Message#decorator_class` for PendingMessages is the right call, the wire format is consistent, and the TUI dispatch covers all current types. A few things worth tightening before or after merge:

---

### Minor Issues

**1. `decorator_class` has no `else` branch**
`app/models/pending_message.rb`

Returns `nil` for any unmapped `message_type`. `Message#decorator_class` has the same gap, so this follows precedent — but since PMs are the universal intake queue and #427 will keep adding types, an explicit guard is cheap insurance:
```ruby
else
  raise ArgumentError, "No decorator for PendingMessage message_type: #{message_type.inspect}"
end
```

**2. `broadcast_payload` sends raw `content` alongside `rendered`**
`app/models/pending_message.rb`

Both `"content"` (raw) and `"rendered"` (decorated) travel on every broadcast. For basic mode, decorators intentionally return `nil` for non-user PMs — but the raw content still crosses the wire, undercutting the view-mode boundary. Also doubles payload size for large tool responses. Consider omitting `"content"` when `rendered` is present, or adding a comment explaining why the legacy fallback is still needed.

**3. `broadcast_payload` default arg triggers an AR load**
`app/models/pending_message.rb`

`def broadcast_payload(mode = session.view_mode)` — the default evaluates `session.view_mode` eagerly. `broadcast_created` calls the no-arg form from an `after_create_commit` callback, so if the association isn't preloaded, that's a SELECT per PM creation. Easy fix — pass explicitly:
```ruby
def broadcast_created
  ActionCable.server.broadcast("session_#{session_id}", broadcast_payload(session.view_mode))
end
```

**4. Three Melete decorators duplicate ~90 lines**
`app/decorators/pending_from_melete_{skill,workflow,goal}_decorator.rb`

`render_basic`, `render_verbose`, and `render_debug` are identical across all three — only `KIND` and `render_melete` differ. A shared `PendingFromMeleteDecorator` base would cut each subclass to ~10 lines. The shared spec already treats them as a family; the class hierarchy should match.

**5. Base class docstring is inaccurate about the fallback chain**
`app/decorators/pending_message_decorator.rb`

Says *"verbose, debug, melete, mneme delegate to basic"* but the code does `render_debug → render_verbose → render_basic` (two-step chain). A subclass that only overrides `render_verbose` gets `render_debug` for free — useful behavior, but the comment obscures it.

---

### Test Coverage Gaps

Worth addressing in a follow-up if not in this PR:

- **TUI rendering for new pending roles** (`render_pending_badge_entry`, `render_pending_subagent_entry`, etc.) has zero test coverage — the chat_spec tests only exercise the `add_pending` routing, not how it actually renders
- **`normalize_pending_data`** symbol→string conversion is never exercised with real decorator output (test helper hardcodes string keys)
- **`pending?` / `effective_color` / `effective_response_color`** in `lib/tui/decorators/base_decorator.rb` — the entire pending-state color muting mechanism — is untested
- All decorator specs use `create(:session)` where `build_stubbed(:session)` would avoid unnecessary DB writes

---

### Nits

- `payload.dig("rendered")` with a single key — use `payload["rendered"]` (dig implies nesting that doesn't exist)
- `success != false` coerces `nil` to `true` silently — deserves a `# nil treated as success; only explicit false signals failure` comment
- Double `.compact` in `pending_entry_data` allocates two intermediate arrays for the same purpose — capture once
- `PendingFromMnemeDecorator` deliberately omits `render_mneme` (circular injection prevention) but the reason is nowhere stated — one comment would save a future contributor from "fixing" it
- Consider a checklist comment on `decorator_class` documenting the 3-file update pattern for new types (decorator subclass + model case + TUI when clause)

---
*Review by Ani — multi-agent analysis (5 specialists: rails, security, performance, testing, documentation)*

bonk-moltbot (approved):
🐭✅ Clean Draper-pattern extension for PendingMessages.

Per-message_type decorators mirror the promoted Message counterparts — each PM type gets its own visual treatment with `status: pending` for dimming. The abstract base with render chain (debug → verbose → basic) and opt-in melete/mneme modes is the right pattern. MessageStore filtering on nil renders keeps mode switches clean.

PR #468, opened 2026-04-29
#466 refactor: ShellSession on tmux

### LINKED ISSUE #466: Rewrite ShellSession: PTY+FIFO → tmux wrapper
## What

Replace the current PTY+FIFO ShellSession (~400 lines) with a thin tmux wrapper (~50-80 lines).

### OOM bug
Current `@read_buffer << @pty_stdout.read_nonblock(4096)` grows unbounded until marker is found. Large output (`find /`, `cat huge_file`, verbose builds) → process killed by OOM. Truncation applies AFTER data is in RAM.

### Wrong abstraction
The PTY+FIFO architecture optimizes for **machine-readable output**: structured stdout, separate stderr, integer exit codes. But the LLM is not a program — it is a human-shaped text processor trained on terminal output text. The optimal interface is what a human sees in a terminal.

### Unnecessary complexity
~400 lines of infrastructure solving problems that do not exist for LLM consumers:
- FIFO + reader thread + stderr mutex (stderr separation LLM does not need)
- Sentinel marker generation/parsing (replaced by `wait-for`)
- Exit code capture via `$?` (humans see error text, not exit codes)
- `stty -echo`, `\r\n` stripping, `Errno::EIO` handling
- `exec bash --norc --noprofile` dance
- `recover_shell` with SIGTERM/SIGKILL fallback
- Environment detection via Open3 subprocess

### tmux as shell backend

```ruby
# Creation (once per Anima session)
system("tmux new-session -d -s anima-#{session_id} -c #{cwd} -x 200 -y 50")

# Blocking command execution (per tool call)
uuid = SecureRandom.hex(8)
system("tmux send-keys -t anima-#{id} '#{command} ; tmux wait-for -S done-#{uuid}' Enter")
system("tmux wait-for done-#{uuid}")  # blocks until done
output = `tmux capture-pane -pJ -t anima-#{id} -S -#{lines}`

# CWD (free)
cwd = `tmux display-message -p -t anima-#{id} '#{pane_current_path}'`

# Cleanup
system("tmux kill-session -t anima-#{id}")
```

### Key design points

1. **Login shell once** — user gets their PATH, env, aliases
2. **Blocking via `wait-for`** — no polling, no sentinel markers in output
3. **Text capture** — `capture-pane -pJ` = what a human sees. No parsing needed
4. **OOM-safe** — `history-limit` caps scrollback at kernel level. `capture-pane -S -N` grabs bounded tail
5. **No exit codes** — errors are visible in text
6. **No stderr separation** — merged output IS the human terminal interface
7. **Interrupt** — `send-keys C-c`
8. **Crash recovery** — tmux sessions survive Anima crashes (bonus)

### What gets deleted

- `@read_buffer` unbounded accumulation (OOM source)
- FIFO creation, reader thread, stderr mutex, drain timing
- Sentinel marker generation and parsing
- `stty -echo`, `\r\n` stripping
- `recover_shell` (Ctrl+C + drain + respawn)
- Exit code capture via `$?`
- `Errno::EIO` handling
- `exec bash --norc --noprofile`
- Environment detection via Open3 subprocess

### What stays

- `ShellSession.for_session(session)` API
- `shell_session.run(command)` API
- `shell_session.finalize`
- `at_exit` cleanup
- Mutex for thread safety
- Timeout support
- Bash tool as thin consumer

### Size reduction

Current: ~400 lines → Proposed: ~50-80 lines

## Open questions

1. tmux as hard dependency — fallback for containers without tmux?
2. Output window — how many lines to capture per command? Configurable?
3. Prompt visibility — filter shell prompt or leave it? (Model handles it fine)
4. Git/env detection — keep separate or move into tmux commands?

## References

- Full brainstorm: `thoughts/shared/notes/2026-04-28/tmux-shell-session-rewrite.md`
- Original research: `thoughts/shared/notes/2026-03-08/persistent-shell-session-research.md`
- Original decision: `thoughts/shared/notes/2026-03-08/persistent-shell-decision.md`
- Current implementation: `lib/shell_session.rb` (~400 lines)

hoblin:
Closes #466.

## Summary

Replaces the ~400-LOC PTY+FIFO `ShellSession` with a ~120-LOC tmux wrapper. Closes the OOM bug (`@read_buffer` unbounded accumulation in the old PTY code) by capping scrollback at the tmux server level — `history-limit 5000` lines bounds memory at the kernel rather than in Ruby.

The interface for callers (`for_session`, `release`, `#run`, `#pwd`, `#alive?`, `#finalize`) is preserved. `Tools::Bash` now returns the rendered terminal text (what a human would see) instead of separate stdout/stderr/exit_code/env_summary fields, matching the issue's "human-shaped text processor" framing.

## What's gone

- PTY + FIFO + stderr reader thread + sentinel-marker parsing
- `@read_buffer` unbounded accumulation (OOM source)
- `recover_shell` SIGTERM/SIGKILL dance
- `EnvironmentSnapshot` + Open3 git/file detection (the shell prompt now serves as live env telemetry for the agent)
- `@sessions_by_id` in-process registry + `Monitor` + `at_exit` cleanup — tmux is the single source of truth, sessions survive Anima crashes
- `Tools::Bash` `mode` parameter (relied on exit codes that are now gone)
- `project_files_whitelist` / `project_files_max_depth` settings
- ~280 LOC of test scaffolding around the above

## What's new

- One tmux session per `Session` (`anima-shell-{session_id}`), torn down only via explicit `release` / `finalize`
- Sub-agents inherit cwd from parent's tmux pane via `display-message` — dynamic lookup at child shell creation, not snapshot-at-spawn
- `tmux wait-for` channel for command completion (with manual kill of the wait-for child after Ctrl+C, since interactive bash discards the rest of its line on SIGINT)
- `clear-history` + shell `clear` (which writes `\e[3J` on modern terminfo) wipes both visible pane and scrollback so each capture contains exactly the new command's output
- `capture-pane -pJ -S -` returns merged terminal text — `-J` joins wrapped lines so long single-line outputs come back whole
- Pager-disabling env vars (`PAGER=cat`, `GIT_PAGER=cat`, `LESS=-eFRX`, etc.) injected via `send-keys export ...` after the shell starts, so they beat any `export PAGER=less` in the user's `~/.zshrc` etc. — without this, `gh`/`git`/`man` open `less` and block the pane forever
- README install section lists `tmux` as a hard runtime requirement

## Test plan

- [x] `bundle exec rspec spec/lib/shell_session_spec.rb` — passes
- [x] `bundle exec rspec spec/lib/tools/bash_spec.rb` — passes
- [x] `bundle exec rspec spec/integration/session_happy_path_smoke_spec.rb` — passes (real tmux + real `gh`)
- [x] `bundle exec rspec spec/lib/tools/spawn_subagent_spec.rb spec/lib/tools/spawn_specialist_spec.rb spec/lib/anima/settings_spec.rb spec/jobs/ spec/lib/providers/anthropic_spec.rb` — all pass
- [x] `bundle exec standardrb` clean
- [ ] Manual smoke (real Anima session) — pending review

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,8 @@ Two more subsystems are designed but not yet implemented: **Thymos** (a hormonal
 
 Anima is a Rails app distributed as a gem, following Unix philosophy: immutable program separate from mutable data.
 
+**Requirements:** Ruby 3.4+, `tmux` (used as the persistent shell backend for the bash tool), `gh` (GitHub CLI, for issue/PR tools).
+
 ```bash
 gem install anima-core       # Install the Rails app as a gem
 anima install                # Create ~/.anima/, set up databases, start brain as systemd service
--- a/lib/anima/settings.rb
+++ b/lib/anima/settings.rb
@@ -194,16 +194,6 @@ module Anima
       # @return [String] absolute path
       def soul_path = get("paths", "soul")
 
-      # ─── Environment ──────────────────────────────────────────────
-
-      # Filenames to scan for in the working directory.
-      # @return [Array<String>]
-      def project_files_whitelist = get("environment", "project_files")
-
-      # Maximum directory depth for project file scanning.
-      # @return [Integer]
-      def project_files_max_depth = get("environment", "project_files_max_depth")
-
       # ─── GitHub ─────────────────────────────────────────────────────
 
       # Repository for feature requests (+owner/repo+ format).
--- a/lib/shell_session.rb
+++ b/lib/shell_session.rb
@@ -34,7 +34,6 @@ class ShellSession
   PANE_WIDTH = 200
   PANE_HEIGHT = 50
   HISTORY_LIMIT = 5_000
-  IDLE_POLL_INTERVAL = 0.1
 
   # Env vars that disable interactive pagers and credential prompts in
   # the shell. Without these, tools like +gh+, +git+, +man+, +journalctl+
@@ -218,29 +217,27 @@ class ShellSession
     line = "#{exports}; tmux wait-for -S init-#{uuid}"
     system("tmux", "send-keys", "-t", @target, line, "Enter", out: File::NULL, err: File::NULL)
     pid = Process.spawn("tmux", "wait-for", "init-#{uuid}", out: File::NULL, err: File::NULL)
-    deadline = monotonic_now + 5
-    loop do
-      _, status = Process.wait2(pid, Process::WNOHANG)
-      break if status
-      if monotonic_now > deadline
-        begin
-          Process.kill("TERM", pid)
-        rescue
-          Errno::ESRCH
-        end
-        begin
-          Process.wait(pid)
-        rescue
-          Errno::ECHILD
-        end
-        raise "tmux session #{@target} init timed out"
-      end
-      sleep IDLE_POLL_INTERVAL
+    waiter = Process.detach(pid)
+    return if waiter.join(5)
+
+    begin
+      Process.kill("TERM", pid)
+    rescue
+      Errno::ESRCH
     end
+    waiter.join
+    raise "tmux session #{@target} init timed out"
   end
 
-  # Blocks on +tmux wait-for+, polling for interrupt and timeout. On
-  # cancel we send +C-c+ to abort the running command, then kill the
+  # Blocks until the +tmux wait-for+ child exits (= the bash command in
+  # the pane finished and ran +tmux wait-for -S+), the deadline expires,
+  # or the user requests an interrupt.
+  #
+  # No polling: {Process.detach} returns a Thread that waits on the
+  # child, and {Thread#join} blocks until either the thread finishes or
+  # the timeout fires — returning immediately when the child exits.
+  #
+  # On cancel we send +C-c+ to abort the running command, then kill the
   # wait-for child directly — interactive bash discards the rest of the
   # input line on SIGINT, so the trailing +tmux wait-for -S+ never fires
   # and we can't rely on natural signaling.
@@ -248,41 +245,34 @@ class ShellSession
   # @return [Symbol] +:done+, +:interrupted+, or +:timeout+
   def wait_for_completion(uuid, timeout, interrupt_check)
     pid = Process.spawn("tmux", "wait-for", "done-#{uuid}", out: File::NULL, err: File::NULL)
+    waiter = Process.detach(pid)
     deadline = monotonic_now + timeout
-    poll_interval = interrupt_check ? Anima::Settings.interrupt_check_interval : IDLE_POLL_INTERVAL
+    interrupt_interval = interrupt_check ? Anima::Settings.interrupt_check_interval : timeout
 
     loop do
-      _, status = Process.wait2(pid, Process::WNOHANG)
-      return :done if status
+      remaining = deadline - monotonic_now
+      return cancel_command(pid, waiter, :timeout) if remaining <= 0
 
-      if interrupt_check&.call
-        cancel_command(pid)
-        return :interrupted
-      end
+      # Block until child exits, the next interrupt-check tick, or the deadline.
+      slice = [remaining, interrupt_interval].min
+      return :done if waiter.join(slice)
 
-      if monotonic_now > deadline
-        cancel_command(pid)
-        return :timeout
-      end
-
-      sleep poll_interval
+      return cancel_command(pid, waiter, :interrupted) if interrupt_check&.call
     end
   end
 
-  # Sends Ctrl+C to abort the running command, then reaps the
-  # wait-for child. Idempotent against already-dead processes.
-  def cancel_command(wait_for_pid)
+  # Sends Ctrl+C to abort the running command, kills the wait-for
+  # child, and waits for the detach thread to reap it. Returns +state+
+  # so call sites can inline +return cancel_command(...)+.
+  def cancel_command(pid, waiter, state)
     send_ctrl_c
     begin
-      Process.kill("TERM", wait_for_pid)
-    rescue Errno::ESRCH
-      # Already exited — wait-for unblocked between our last poll and now
+      Process.kill("TERM", pid)
+    rescue
+      Errno::ESRCH
     end
-    begin
-      Process.wait(wait_for_pid)
-    rescue Errno::ECHILD
-      # Already reaped
-    end
+    waiter.join
+    state
   end
 
   def send_ctrl_c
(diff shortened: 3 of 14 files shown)


hoblin:
## PR #468 Review — ShellSession PTY+FIFO → tmux

**Verdict: REQUEST_CHANGES** (two issues worth addressing before merge)

### Major

**1. `capture_output` silently returns `"OK"` on session death**
`lib/shell_session.rb` — `capture_output`

`Open3.capture2` status is discarded (`raw, _ = ...`). If the tmux session dies between `wait_for_completion` and the capture, `capture-pane` exits non-zero, `raw` is `""`, and the caller gets `{output: "OK"}` — indistinguishable from a silent command that succeeded. The LLM has no way to know the session crashed.

```ruby
# current
raw, _ = Open3.capture2("tmux", "capture-pane", ...)

# suggested
raw, status = Open3.capture2("tmux", "capture-pane", ...)
return { error: "tmux capture-pane failed (session may have died)" } unless status.success?
```

**2. `inject_shell_env` — unbounded hang after SIGTERM**
`lib/shell_session.rb` — `inject_shell_env`

After sending SIGTERM to a wedged `wait-for` child, `waiter.join` is called with no timeout. If the process ignores TERM, `ShellSession.new` hangs forever.

```ruby
# current
Process.kill("TERM", pid) rescue Errno::ESRCH
waiter.join  # ← no timeout

# suggested
Process.kill("TERM", pid) rescue Errno::ESRCH
waiter.join(5)
```

### Minor

**3. Child sessions get `initial_cwd: nil`**
`lib/tools/spawn_subagent.rb:87`, `lib/tools/spawn_specialist.rb:102`

`SpawnSubagent`/`SpawnSpecialist` no longer set `initial_cwd:` on child sessions. Dynamic lookup via `parent_cwd_for` works when the parent tmux session is alive, but the fallback is `session.initial_cwd` which is now nil. The spec uses a double with `initial_cwd: "/"`, masking this in CI.

**4. No orphan sweep**
`lib/shell_session.rb` (absence)

Old code had `at_exit { ShellSession.cleanup_all }`. New code has no equivalent — tmux sessions from crashed Anima processes accumulate indefinitely. A startup sweep of `anima-shell-*` sessions or a note in ops docs would close the gap.

**5. Concurrent `for_session` race**
`lib/shell_session.rb` — `.for_session`

The old `Monitor`-based lock is gone. Two threads calling `for_session` on a cold session could both see `alive?` → false, both call `new-session` + `inject_shell_env`, and corrupt the pane with competing init lines. Low probability in practice but the old test existed for a reason.

**6. VCR cassette changed**
`spec/cassettes/Providers_Anthropic/...raises_ServerError_on_529_overload.yml`

Binary diff in an Anthropic API cassette is unexpected for a shell session rewrite. Worth verifying it wasn't re-recorded against the live API.

### Nit

- `capture_output`: `raw.dup.force_encoding(...)` — `dup` is unnecessary; `force_encoding` mutates the receiver
- `PANE_CWD_FORMAT` is a one-use constant with a rubocop-disable; could be inlined
- `rescue => error` has `# rubocop:disable Lint/RescueException` but bare rescue catches `StandardError`, not `Exception` — the disable comment is targeting the wrong cop
- Removing `mode: "parallel"` silently ignores legacy tool calls; a deprecation log would surface stale prompts

---

*Reviewed by five parallel subagents: convention scout, tmux lifecycle, performance, test quality, documentation.*

hoblin:
## Review feedback resolutions — `0f19929`

### Major
**1. `capture_output` silent "OK" on session death** — Fixed. `capture_output` now returns `nil` when `capture-pane` exits non-zero; `#run` surfaces this as `{error: "tmux capture-pane failed (session may have died)"}`. Spec covers the path by stubbing the failed status.

**2. `inject_shell_env` unbounded post-TERM hang** — Fixed. Extracted `reap_waiter(pid, waiter)` that does `TERM → join(WAITER_KILL_GRACE) → KILL → join`. Both `inject_shell_env` and `cancel_command` use it now.

### Minor
**3. Child sessions get `initial_cwd: nil`** — Fixed. `SpawnSubagent`/`SpawnSpecialist` snapshot `ShellSession.cwd_via_tmux(parent_id) || parent.initial_cwd` at `Session.create!`. Dynamic lookup still wins when the parent's tmux is alive; the snapshot just makes the fallback path useful (e.g. across an Anima restart). Specs added in both spawn specs.

**4. No orphan sweep** — Declined. Per the April brainstorm (`thoughts/shared/notes/2026-04-28/tmux-shell-session-rewrite.md`), crash recovery is "bonus, not requirement" — SQLite is the source of truth for conversation state, and an orphan tmux session with no matching `Session` row is harmless until the next reboot. Adding a startup sweep extends scope beyond what was decided.

**5. Concurrent `for_session` race** — Fixed with a class-level `INIT_MUTEX` around `ensure_session`. Single global mutex (vs. per-`session_id` map) is sufficient because the cold path is fast and the warm path takes the early `alive?` return without contention.

**6. VCR cassette change** — Declined. Verified the diff: only the recorded request body changed (Bash tool's input schema lost `mode`, and a system-prompt block re-wrapped). The 529 response body is byte-identical to the pre-PR cassette, and the test still asserts `ServerError`. This is a re-record, not a re-recording against the live API.

### Nits
- `raw.dup.force_encoding(...)` — Fixed. Dropped the `dup`.
- `PANE_CWD_FORMAT` constant — Declined. The named constant + adjacent comment + rubocop disable together document why this string is single-quoted (tmux performs the `#{...}` substitution server-side). Inlining would lose that.
- Wrong rubocop disable on bare `rescue` — Fixed. `Lint/RescueException` doesn't apply to bare rescue (which catches `StandardError`); standardrb wasn't actually flagging anything here. Replaced the disable directive with a comment that documents the LLM-tool-call boundary intent.
- `mode: "parallel"` deprecation log — Declined. Old in-flight tool calls aren't a real concern in this codebase (single-instance Anima, no rolling deploy), and the cassette update covers the only test surface where the legacy field appeared.

bonk-moltbot (approved):
🐭✅ PTY+FIFO → tmux. Beautiful simplification.

~400 LOC → ~120 LOC. The OOM bug (`@read_buffer` unbounded growth) is solved architecturally — tmux `history-limit` caps memory at the kernel level instead of relying on Ruby-side discipline.

Review feedback was thorough and well-addressed:
- `capture_output` now returns nil on dead sessions (was silently returning "OK")
- `reap_waiter` TERM → join(5) → KILL guarantees no hangs
- `INIT_MUTEX` closes the concurrent cold-start race
- Child `initial_cwd` properly snapshots from parent's tmux pane

Key design wins:
- tmux is the source of truth — ShellSession is a disposable handle. Sessions survive Anima crashes
- `capture-pane -pJ` gives "what a human would see" — the right abstraction for an LLM
- Pager-disabling env vars injected *after* rcfiles, beating user `export PAGER=less`
- `wait-for` + `C-c` interrupt pattern handles the interactive bash SIGINT edge case correctly

CI all green across Ruby 3.2/3.3/3.4.

Issue #11, opened 2026-03-07
Session and event persistence


hoblin:
## Problem to solve

Events flow in memory (#10) but disappear when the TUI closes. Users lose their entire conversation on exit. Without persistence, Anima can't resume conversations, and there's no data for the viewport to assemble context from.

## Solution

Persist events to SQLite. Introduce Session model that owns an ordered event stream. Conversation survives TUI restart.

### Models
- `Session` — owns an ordered event stream
- `Event` — polymorphic type, payload (JSON), session_id, timestamp, position

### Event persistence subscriber
- Listens on the event bus, writes events to SQLite as they flow

### Context assembly
- Load events from session, build messages array for LLM
- Replaces the in-memory array from #9

### TUI integration
- TUI subscribes to event bus for real-time updates
- `anima tui` resumes last session or creates new one
- `Ctrl+a` > `n` creates a new session

### Session lifecycle
- Sessions are created on first message or via `Ctrl+a` > `n`
- Reopening TUI resumes the last active session

## Verification

```ruby
# Have a conversation, quit, reopen TUI — conversation is there
# In Rails console:
Session.last.events.count  # => shows persisted events
```

PR #470, opened 2026-04-29
#469 fix: stop MCP stdio process leak

### LINKED ISSUE #469: MCP server processes leak: brave-search-mcp-server never cleaned up
## Problem to solve

MCP server processes spawned for tool calls are never terminated when the conversation or session ends. Each Brave Search tool call starts an `npm exec @brave/brave-search-mcp-server` which spawns a `node` child — neither gets killed after the call completes.

On a running Anima instance after ~24h of normal use:

- **699 brave-search-mcp-server processes** (and growing)
- **~66 GB RSS** reported (shared pages inflate this, real unique ~33 GB)
- **562 of 699 completely idle** (0% CPU) — zombie servers sitting in memory
- **341 direct children of solid-queue supervisor** — the rest are orphaned npm wrappers

This is the OOM source that has been killing long-running Anima sessions.

## Solution

MCP client must track spawned server processes and terminate them when:
1. The tool call completes (if server-per-call model)
2. The conversation/session ends (if server-per-session model)
3. Anima shuts down (at_exit sweep)

## Diagnostics

```bash
# Count zombie MCP servers
ps aux --no-headers | grep -c 'brave-search-mcp-server'

# See total RSS
ps aux --no-headers | grep 'brave-search-mcp-server' | awk '{sum+=$6} END {printf "%.0f MB\n", sum/1024}'

# Kill them all (immediate relief)
pkill -f 'brave-search-mcp-server'
```

## Why this wasn't caught earlier

The per-process RSS is small (~95 MB each) — no single process stands out in top/htop. The leak is only visible when aggregating by command name across hundreds of processes.

hoblin:
Closes #469.

## Summary

Two compounding bugs caused `brave-search-mcp-server` processes to accumulate (699 procs / ~66 GB RSS after 24h, the OOM source for long-running Anima sessions):

1. **`Mcp::ClientManager` was rebuilt on every job.** `Tools::Registry.build` ran on every `DrainJob` and `ToolExecutionJob`, constructing a fresh manager that spawned a fresh `StdioTransport` per configured server. The class-level `@instances` registry tracked transports for `at_exit` cleanup but was never unregistered during normal operation — so each job leaked another `npx → node` subtree. `Mcp::ClientManager.shared` now memoizes the manager process-wide, so each configured server is connected exactly once per worker lifetime, and tool wrappers are cached after the first `tools/list`. This matches the architectural intent recorded in `thoughts/shared/notes/2026-03-14/mcp-client-integration-research.md` (decision #11: lazy-spawn, persistent, respawn on crash).

2. **No process-group isolation at spawn.** `Open3.popen2` ran without `pgroup: true`, so `npx`/`npm` wrappers shared the worker's process group. Even when `terminate_process` did fire (e.g. via `at_exit`), only the wrapper PID got the signal — its `node` grandchild orphaned to PID 1. `StdioTransport` now spawns with `pgroup: true` and signals the whole group via `Process.kill("TERM"/"KILL", -pid)`. Wrapper, `node` child, and any descendants die together.

The existing `at_exit { Mcp::StdioTransport.cleanup_all }` hook handles teardown on worker shutdown; with the bounded instance count it's now meaningful.

## Test plan

New regression coverage (all green):

- [x] `StdioTransport` spawns the server as its own process-group leader (`pgid == pid`)
- [x] Shutting down the wrapper terminates a forked grandchild (the `npx → node` analogue)
- [x] Repeated `register_tools` calls reuse cached transports — `Open3.popen2` called once across N jobs
- [x] Tool catalogs are fetched once per server, not once per job
- [x] Cached tool wrappers attach to every fresh `Tools::Registry`
- [x] `Mcp::ClientManager.shared` returns the same instance across calls

```
bundle exec rspec spec/lib/mcp/ spec/lib/tools/registry_spec.rb spec/lib/tools/mcp_tool_spec.rb
# 131 examples, 0 failures
```

`bundle exec standardrb` clean on every touched file.

## Breaking changes

None. Public surface (`Tools::Registry.build`, MCP tool dispatch, `mcp.toml` semantics) is unchanged. Sub-agents continue to share the MCP client pool — they live in the same worker process, which matches the persistent-per-worker model.

### DIFF
--- a/lib/mcp/client_manager.rb
+++ b/lib/mcp/client_manager.rb
@@ -3,124 +3,68 @@
 require "mcp"
 
 module Mcp
-  # Manages MCP client connections and registers their tools with
-  # {Tools::Registry}. Each configured server (HTTP or stdio) gets a
-  # dedicated {MCP::Client} instance that persists for the worker
-  # process's lifetime, in line with the MCP integration research note
-  # (decision #11): stdio servers spawn lazily on first use and stay
-  # alive across tool calls.
+  # Connects to MCP servers and registers their tools with
+  # {Tools::Registry}. Each configured server (HTTP or stdio) is
+  # connected on first use and the resulting tool wrappers are cached
+  # for the worker's lifetime — so {Tools::Registry.build}, which fires
+  # per job, reuses the same MCP clients instead of respawning a fresh
+  # subprocess per server per job (issue #469).
   #
-  # The class-level {.shared} accessor returns a single instance per
-  # process. {Tools::Registry.build} is called per job, but it pulls
-  # tools from the shared manager, so each configured server is connected
-  # exactly once per worker — not once per job — preventing the
-  # subprocess accumulation reported in issue #469.
+  # Spawned stdio server processes are reaped on worker exit by
+  # {Mcp::StdioTransport.cleanup_all} (registered as an at_exit hook).
   #
   # Connection failures are logged and skipped — a misconfigured or
   # unavailable server does not prevent other servers or built-in tools
-  # from working. Warnings collected during the initial load are returned
-  # from the first {#register_tools} call so the caller can surface them
-  # as system messages; subsequent calls return an empty array, since the
-  # same warnings would otherwise be re-emitted on every job.
+  # from working.
   #
-  # @example Production use
+  # @example
   #   Mcp::ClientManager.shared.register_tools(registry)
-  #
-  # @example Test use with injected config
-  #   manager = Mcp::ClientManager.new(config: fake_config)
-  #   manager.register_tools(registry)
   class ClientManager
-    class << self
-      # Process-wide shared instance. Lazily constructed on first call.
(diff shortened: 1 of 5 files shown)


bonk-moltbot (approved):
🐭✅ The real OOM killer, caught and buried.

Two compounding bugs, both fixed cleanly:

**1. `ClientManager` rebuilt on every job** — `.shared` memoizes process-wide. Each stdio server spawns once per worker lifetime. Tool wrappers cached after first `tools/list`. Test proves `Open3.popen2` called exactly once across N `register_tools` calls.

**2. No process-group isolation** — `pgroup: true` on spawn + `Process.kill('TERM', -pid)` kills the entire subtree. The `npx → node` grandchild leak is architecturally impossible now. The grandchild-reaping spec (spawns a real forking script, asserts the grandchild dies on terminate) is excellent — this is the kind of test that prevents regressions.

699 processes / 66 GB → 1 process per configured server. 🎯

PR #484, opened 2026-04-30
#482 fix: trim trailing blank lines from tmux capture-pane output

### LINKED ISSUE #482: Trim trailing blank lines from bash tool responses captured via tmux
## Problem to solve

After the `bash` tool was rewritten to use a tmux backend (#466), some tool responses come back with trailing blank lines. These don't carry information — they're padding lines from `tmux capture-pane` filling the visible pane height — but they:

- Take up screen space in the TUI for no reason
- Cost tokens in the LLM's context window
- Make the visible end of output look like the command might still be running

Under the previous PTY+FIFO implementation the captured output ended at the last meaningful line, so this is a regression introduced by the tmux switch.

## Solution

Strip trailing blank lines from the captured output before returning it. Preserve trailing newlines on the final non-blank line — only collapse runs of blank lines at the very end.

A one-line change to `ShellSession#capture_output` (`lib/shell_session.rb:342-346`):

```ruby
def capture_output
  raw, status = Open3.capture2("tmux", "capture-pane", "-pJ", "-t", @target, "-S", "-", err: File::NULL)
  return nil unless status.success?
  output = truncate(raw.force_encoding("UTF-8").scrub)
  output = output.sub(/\n+\z/, "\n")  # ← collapse trailing blank lines
  output.strip.empty? ? EMPTY_OUTPUT_PLACEHOLDER : output
end
```

## Why this approach

- `tmux capture-pane -pJ -S -` returns the whole scrollback; tmux pads with blank lines to fill the visible pane geometry. The padding is purely a rendering artifact, not part of the command's output.
- Trimming after `truncate` (not before) ensures the byte-budget contract still holds — `truncate` capped the raw, and the trim can only shrink it further.
- Keeping a single trailing `\n` matches POSIX command-output convention and avoids surprising callers that expect output to end with a newline.
- The existing `output.strip.empty? ? EMPTY_OUTPUT_PLACEHOLDER : output` branch still works correctly — an output that was *only* blank lines becomes `""` after the trim, then maps to `"OK"`.

## Tests

Add coverage in `spec/lib/shell_session_spec.rb`:

- Output ending in multiple blank lines → single trailing newline preserved
- Output ending in a non-blank line with no trailing newline → unchanged
- Output that is only blank lines → `EMPTY_OUTPUT_PLACEHOLDER` ("OK")
- Output exactly at the byte cap with trailing blank padding → trim happens, truncation notice is preserved

## Why this matters

Every blank line is ~1 token wasted across thousands of tool calls per session. More importantly, an agent reading its own captured output may treat trailing blanks as semantically meaningful (e.g. "the command produced empty trailing output, did something fail?") when in fact they're just terminal geometry. The fix is small and the user-visible improvement is immediate.

## Related

- #466 — Rewrite ShellSession: PTY+FIFO → tmux wrapper

hoblin:
## Summary

`tmux capture-pane -pJ -S -` returns the full scrollback padded out to `PANE_HEIGHT` with empty rows. Without trimming, every byte of that padding ends up in the LLM's context for no reason — and can blow the truncation budget for legitimately small output.

The previous PTY+FIFO `ShellSession` explicitly filtered blank lines at read time (`output << line if line && !line.strip.empty?`); the tmux rewrite (#466) lost that behaviour. This restores it at the capture boundary.

## Why this matters (measured)

Spot-check against this Anima's production DB, last 15 bash tool responses:

| Metric | Value |
|---|---|
| Total bytes | 17,191 |
| After trim | 14,093 |
| **Saved** | **3,098 bytes (18.0%)** |
| Avg per response | 207 bytes of pure padding |

Some individual responses were 30–42% padding. Across a long session that's substantial token-budget reclaim for zero information loss.

## Changes

- `lib/shell_session.rb` — `capture_output` now collapses a trailing run of `\n` + whitespace into a single trailing newline before passing to `truncate`. Order matters: trim must happen **before** truncate so pane padding doesn't blow the byte cap and trigger a false `[Truncated:]` notice.
- `.dup` the captured string before `force_encoding` — caught while writing the regression test (frozen-string literal raised `FrozenError` in `rescue`-swallowed code path).
- Doc comment on `capture_output` explaining why the trim exists and why the order matters.

## Test plan

- [x] `bundle exec rspec spec/lib/shell_session_spec.rb` — 41/41 green (added 5 new examples in `context "trailing blank lines"`).
- [x] Real-tmux integration tests confirm production happy-path behaviour.
- [x] Mocked-Open3 tests pin the trim regex against synthetic captures with the exact shape produced by tmux (`"line\n   \n   \n"`, not just `"line\n\n\n"`).
- [x] Regression test covers frozen-string capture (e.g. future caller passing in a literal).
- [x] `bundle exec standardrb` clean.
- [x] `bundle exec reek lib/shell_session.rb` — same warnings as `main`, no new smells.

## Closes

Closes #482

## Related

- #466 — Rewrite ShellSession: PTY+FIFO → tmux wrapper (the change that introduced this regression)

### DIFF
--- a/lib/shell_session.rb
+++ b/lib/shell_session.rb
@@ -334,11 +334,14 @@ class ShellSession
   # The +-J+ flag joins terminal-wrapped lines so a long single-line
   # output comes back whole.
   #
-  # Trailing blank lines are collapsed to a single newline before
-  # truncation. +tmux capture-pane -S -+ pads the captured scrollback
-  # with empty rows to fill +PANE_HEIGHT+, and that padding is purely a
-  # rendering artifact — every byte of it would otherwise count against
-  # the truncation budget and end up in the LLM's context for no reason.
+  # Trailing whitespace-only rows are collapsed to a single newline
+  # before truncation. +tmux capture-pane -S -+ pads the captured
+  # scrollback with empty rows to fill the pane height, and each row is
+  # padded with spaces to the pane width — so the trailing artifact
+  # looks like +"\n   \n   \n"+, not just +"\n\n\n"+. That is why the
+  # regex matches +\s*+, not +\n*+. The padding is purely a rendering
+  # artifact: every byte of it would otherwise count against the
+  # truncation budget and end up in the LLM's context for no reason.
   # Trimming before {#truncate} keeps the byte cap honest: a small
   # command followed by 50 lines of pane padding no longer registers as
   # "output exceeded N bytes."
@@ -351,6 +354,8 @@ class ShellSession
   def capture_output
     raw, status = Open3.capture2("tmux", "capture-pane", "-pJ", "-t", @target, "-S", "-", err: File::NULL)
     return nil unless status.success?
+    # +.dup+: +force_encoding+ mutates in place; defends against frozen callers (e.g. test mocks
+    # passing string literals when +# frozen_string_literal: true+ is set).
     cleaned = raw.dup.force_encoding("UTF-8").scrub.sub(/\n\s*\z/, "\n")
     output = truncate(cleaned)
     output.strip.empty? ? EMPTY_OUTPUT_PLACEHOLDER : output
--- a/lib/shell_session.rb
+++ b/lib/shell_session.rb
@@ -333,6 +333,16 @@ class ShellSession
   # trailing prompt — nothing leaked from the previous pane state.
   # The +-J+ flag joins terminal-wrapped lines so a long single-line
   # output comes back whole.
+  #
+  # Trailing blank lines are collapsed to a single newline before
+  # truncation. +tmux capture-pane -S -+ pads the captured scrollback
+  # with empty rows to fill +PANE_HEIGHT+, and that padding is purely a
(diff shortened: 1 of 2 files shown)


bonk-moltbot (approved):
🐭✅ Clean regression fix from the tmux rewrite.

18% token savings measured on production data — 207 bytes of pure padding per response. The regex `\n\s*\z` is correct: tmux pads each row with spaces to pane width, so the artifact is `"\n   \n   \n"`, not just `"\n\n\n"`. Tests pin both shapes.

Good catches:
- **Trim before truncate** — order matters or pane padding blows the byte cap and triggers false `[Truncated:]` notices
- **`.dup` before `force_encoding`** — frozen string literal defense, caught while writing tests (would have been a silent swallow in the `rescue` path)
- **Frozen capture test** — relies on the file's `frozen_string_literal: true` pragma, which is a nice self-documenting assertion

The old PTY+FIFO code had `!line.strip.empty?` filtering — this restores equivalent behavior at the capture boundary instead of line-by-line.

Issue #480, opened 2026-04-30
Research spike: source of `from_*` tool calls in the main loop


hoblin:
## Problem to solve

Aoide's main loop occasionally surfaces tool calls named after sister/sub-agent message channels — e.g. `from_melete_goal`, `from_zero-width-sleuth`, `from_whitespace-scout`. These names are not registered as tools; they are the prefixes the system uses when delivering messages *to* the agent. Calling them produces:

```
{"error":"Tools::UnknownToolError: Unknown tool: from_<name>"}
```

…polluting the conversation and wasting tokens. Today this is a known papercut that requires the human to nudge the agent ("these are not tools").

We do not currently know whether the offending tool calls originate from the LLM itself or from somewhere inside Anima.

## Two hypotheses

1. **LLM hallucination.** The model sees prior `from_*` blocks in the conversation context and pattern-matches them as callable tools. Anthropic's endpoint returns them in `tool_use` blocks because the model decided to invoke them.
2. **Internal leak.** Some part of Anima's pipeline (event subscriber, tool-execution dispatcher, registry assembly) generates these calls when it shouldn't — they never come from the LLM at all.

The fix differs significantly between the two:

- **If hallucination:** filter `from_*` tool_use blocks out of the dispatch path (`lib/events/subscribers/llm_response_handler.rb#normalize_tool_uses`) and pretend the LLM never called them. No error round-trip.
- **If internal leak:** find the source and stop emitting them.

## Research approach

Wire up debug logging for Aoide's main loop the same way the muses are instrumented today (`lib/melete/runner.rb:132-133` logs system prompt and user message at debug level). For Aoide we additionally need:

- Raw response payload from the Anthropic endpoint, including `content` blocks (text + tool_use)
- The list of tool_use blocks as returned by the API, before any normalization
- The tool name and ID of every dispatched tool execution, so we can correlate "what came in" vs "what got called"

Enable this in dev environment only.

## Acceptance criteria

- [ ] After a session that exhibits the bug, the logs contain enough information to determine whether the `from_*` tool_use block was present in the API response or was synthesized later.
- [ ] Decision recorded in this issue (or a follow-up): **hallucination** vs **internal leak**.
- [ ] Follow-up implementation issue opened with the appropriate fix.

## Suspected choke points (starting hints, not prescribed)

[...]

PR #445, opened 2026-04-11
#436 refactor: the Three Muses — Melete, Mneme, Aoide

hoblin:
Closes #436.

The brainstorm note (`thoughts/shared/notes/2026-04-11/three-muses-melete-aoide-mneme.md`) described five things: rename AnalyticalBrain → Melete, rewrite Melete's and Mneme's system prompts in their own voices, add the `from_` phantom-tool prefix convention, rename Aoide's memory tools, and add a Sisters block to the main system prompt. Everything in that note landed. A handful of sidequests landed with it — worth a separate listing because none of them are in the brainstorm and the PR description shouldn't ask reviewers to spot them mid-diff.

## Slices in the brainstorm
1. `AnalyticalBrain → Melete` mechanical rename (module, job, specs, log paths, cassette dirs)
2. `[analytical_brain] → [melete]` config section + Settings accessors
3. `from_` prefix convention for phantom tool names (`recall_memory → from_mneme`; skill/workflow/goal → `from_melete`; `subagent_message → from_{nickname}` dynamic via lambda)
4. `Tools::Recall` / `Tools::Remember` → `Tools::SearchMessages` / `Tools::ViewMessages`
5. Sisters block added to Aoide's main-session system prompt (after SOUL, before snapshots)
6. Melete's system prompt rewritten in her voice — role before rules, tax framing for irrelevant activation
7. Mneme's L1 + L2 system prompts rewritten in her voice (same tax framing)

## Sidequests the brainstorm did not describe
Landed during implementation once the new mental model clicked. Most of them fell out of a single realisation: if Melete's activations ride the viewport as `from_melete` phantom pairs and naturally evict, then "active" is just **what's in the viewport right now** — no separate state, no deactivation, no sync.

- **Dropped `active_skills` / `active_workflow` DB columns** (migration `20260411172926_remove_active_skills_and_workflow_from_sessions`). They were maintained by AR callbacks to mirror viewport contents. Now derived directly.
- **`Session#active_skills` / `#active_workflow` are methods** — union of pending `PendingMessage` records (source_type `skill`/`workflow`) and in-viewport `from_melete` phantom pairs, read off the viewport on demand.
- **Deleted `DeactivateSkill` / `DeactivateWorkflow` tools** and the paired `Session#deactivate_skill` / `#deactivate_workflow` methods. Deactivation is no longer a thing Melete can do — a skill rides the viewport until it naturally evicts.
[...]

PR #485, opened 2026-04-30
#480 chore: instrument Aoide main loop with dev-only debug logging

hoblin:
## Summary

Wires up dev-only debug logging into `Events::Subscribers::LLMResponseHandler` so we can pin down whether spurious `from_*` tool calls (e.g. `from_melete_goal`, `from_zero-width-sleuth` → `Tools::UnknownToolError`) originate from the LLM (hallucination) or from somewhere inside Anima (internal leak).

This is the **research-spike data-gathering step** of #480 — not a fix. The decision (hallucination vs internal leak) and the follow-up implementation issue come after a session that exhibits the bug is run against the new logs.

## What's logged (at the point the API response enters the main loop)

1. **Raw Anthropic API response payload** — full structure, encoded as TOON, debug level. Captured before `normalize_tool_uses` runs.
2. **Raw `tool_use` blocks** — debug level, TOON-encoded, shown pre-normalization so missing `id`s look missing. If a `from_melete_goal` block is here, the LLM produced it; if it's only in the dispatched list, something else synthesized it.
3. **Each dispatched tool execution** — info level, one line per call, with name + id.

Plus a one-line info summary at session boundary (`response received (N block(s), M tool_use)`) so a reader can scan the timeline without `--debug`.

Payloads are encoded with `Toon.encode` (already used by `lib/tui/screens/chat.rb` and the tool decorators) — same project convention, ~30% fewer lines than `JSON.pretty_generate`, lossless round-trip.

## Where it lives

- `lib/aoide.rb` — new module-level logger, mirrors `Melete.logger` / `Mneme.logger` byte-for-byte: dev-only `Logger.new("log/aoide.log")`, returns `Logger.new(File::NULL)` outside `Rails.env.development?` so call sites stay unconditional.
- `lib/events/subscribers/llm_response_handler.rb` — three log points: `log_raw_response` at the top of `#emit`, info line in `dispatch_tool_executions`, plus `def log = Aoide.logger`.
- `CLAUDE.md` — adds `tail -f log/aoide.log` next to the existing Melete tip.

## How to use the logs

After this lands, run a session in dev that's likely to surface the bug (it tends to happen when sub-agents are messaging in or skills/goals are being injected). Then:

```
tail -f log/aoide.log | grep -E "raw tool_use blocks|dispatching"
```

If a `dispatching tool=from_*` line appears whose tool name is also present in the matching `raw tool_use blocks:` payload, the LLM hallucinated it. If the `from_*` name is absent from the raw blocks but appears in dispatch, something inside Anima is synthesizing the tool_use after the API response.

## Architectural prior

The codebase review for this spike (PR #445 + drain-principles note + `PendingMessage#promote_as_phantom_pair!`) already favours the hallucination hypothesis: phantom `from_*` pairs are assembled inbound by `PendingMessage#promote!` as message-array entries, and the outgoing tool registry (`Tools::Registry.tool_classes_for`) is built exclusively from real `Tool` subclasses — none of which start with `from_`. So an internal leak is architecturally implausible under the current design. The logging confirms or refutes that without us having to assume.

## Smoke test (against dev brain on 42135)

Restarted the brain after this branch was on disk, said hi, read `log/aoide.log` — got:

```
[10:38:34.813] INFO  session=15 — response received (1 block(s), 0 tool_use)
[10:38:34.814] DEBUG  session=15 raw response:
model: claude-opus-4-6
id: msg_01DGMeXArS2ZQMNagL4HH4AU
type: message
role: assistant
content[1]{type,text}:
  text,"…"
…
[10:38:34.814] DEBUG  session=15 raw tool_use blocks:
[0]:
```

Three signals firing as designed; TOON encoding reads cleanly.

## Boy Scout

While in the handler, factored `session.id`, `tool_use["id"]`, and `tool_use["name"]` reads in `dispatch_tool_executions` and `persist_tool_call` to local variables — reek was flagging the duplicate method calls.

## Test plan

- [x] `bundle exec rspec spec/lib/events/subscribers/llm_response_handler_spec.rb` — 13 examples, 0 failures (5 new, all asserting the logger receives the right calls)
- [x] `bundle exec rspec spec/jobs/drain_job_spec.rb` — 11 examples, 0 failures (sanity check on the upstream emitter)
- [x] `bundle exec standardrb` — clean on changed files
- [x] `bundle exec reek lib/events/subscribers/llm_response_handler.rb` — only 2 pre-existing FeatureEnvy warnings remain (inherent iteration shape; not introduced by this PR and not obviously improved by extraction)
- [x] Manual: dev brain smoke test (above) — `log/aoide.log` populates correctly on a benign greeting.
- [ ] Real-session: wait for a session that exhibits the `from_*` bug, read the logs, decide hallucination vs internal leak, record on #480, open follow-up implementation issue.

## Acceptance criteria from #480

- [x] After a session that exhibits the bug, the logs contain enough information to determine whether the `from_*` tool_use block was present in the API response or synthesized later. *(structurally yes — verified by spec + smoke test, awaiting real-session confirmation)*
- [ ] Decision recorded on #480: hallucination vs internal leak. *(blocked on a real session with the new logs)*
- [ ] Follow-up implementation issue opened with the appropriate fix. *(blocked on the decision above)*

Refs #480

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -94,7 +94,7 @@ Always clean up tmux sessions when done. Use `anima-test` as the session name fo
 **Important:** Use `./exe/anima` (not `bundle exec anima`) to test local code changes. The exe uses `require_relative` so it loads local `lib/` directly. `bundle exec` may load the installed gem version instead.
 
 Melete debug log (dev only): `tail -f log/melete.log`
-Aoide debug log (dev only): `tail -f log/aoide.log` — raw API response, raw tool_use blocks, and dispatched tool name/id from the main loop.
+Aoide debug log (dev only): `tail -f log/aoide.log` — raw API response, raw tool_use blocks (pre-normalization), and dispatched tool name/id. Use to correlate "what came in from the API" against "what got dispatched".
 
 ## Triggering API 400 for smoke testing
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -94,6 +94,7 @@ Always clean up tmux sessions when done. Use `anima-test` as the session name fo
 **Important:** Use `./exe/anima` (not `bundle exec anima`) to test local code changes. The exe uses `require_relative` so it loads local `lib/` directly. `bundle exec` may load the installed gem version instead.
 
 Melete debug log (dev only): `tail -f log/melete.log`
+Aoide debug log (dev only): `tail -f log/aoide.log` — raw API response, raw tool_use blocks, and dispatched tool name/id from the main loop.
 
 ## Triggering API 400 for smoke testing
 
--- a/lib/aoide.rb
+++ b/lib/aoide.rb
@@ -1,9 +1,8 @@
 # frozen_string_literal: true
 
-# Aoide — the muse of voice. The agent's main conversational loop:
-# she takes the system prompt, recent messages, and tool registry, and
-# turns each LLM response into dispatched tool executions. One of the
-# Three Muses: Melete prepares, Aoide performs, Mneme remembers.
+# Aoide — the muse of voice. Turns each LLM response into dispatched
+# tool executions and persisted messages. One of the Three Muses: she
+# performs while Melete prepares the stage and Mneme remembers.
 module Aoide
   # Dev-only logger that writes to log/aoide.log.
   # In non-development environments returns a null logger so
--- a/lib/aoide.rb
+++ b/lib/aoide.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+# Aoide — the muse of voice. The agent's main conversational loop:
+# she takes the system prompt, recent messages, and tool registry, and
+# turns each LLM response into dispatched tool executions. One of the
+# Three Muses: Melete prepares, Aoide performs, Mneme remembers.
+module Aoide
+  # Dev-only logger that writes to log/aoide.log.
+  # In non-development environments returns a null logger so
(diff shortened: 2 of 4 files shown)


bonk-moltbot (approved):
🐭✅ Research spike done right — instrument first, decide later.

**Aoide module** mirrors Melete/Mneme logger pattern byte-for-byte: dev-only file logger, null logger in production, no conditionals at call sites. Clean.

**Three log points at the right seam:**
1. Info summary at response entry — `response received (N block(s), M tool_use)` — scannable timeline
2. Debug: raw response + raw tool_use blocks **pre-normalization** — so missing `id`s look missing, and `from_*` names are traceable to source
3. Info: each dispatched tool name+id — the "what got acted on" counterpart

The correlation workflow is elegant: if `from_melete_goal` appears in raw blocks AND dispatch → LLM hallucinated it. If only in dispatch → internal leak. The architectural prior (phantom `from_*` pairs built inbound by `PendingMessage#promote!`, not in outbound registry) already points at hallucination, but now there's data.

**Boy scout bonus:** local variable extraction in `persist_tool_call` and `dispatch_tool_executions` — fewer repeated hash lookups, cleaner reading. Block form on `log.debug` so `Toon.encode` never runs unless debug level — no perf impact in production.

Test coverage: 5 specs including the "trace a spurious from_* call from raw blocks to dispatch" scenario — directly tests the diagnostic use case.

Issue #481, opened 2026-04-30
TUI pending tool responses leak: some entries persist until restart


hoblin:
## Problem to solve

In the TUI, tool-response PendingMessages occasionally accumulate as visible entries that never disappear. Once stuck, they persist for the lifetime of the TUI process — only a restart clears them. Real `Message` rows do exist in the DB for these tool calls; the duplicates are stale `PendingMessage` entries that never received their removal broadcast.

This is the second known papercut in the pending-message broadcast path. The first — pending entries being styled like user messages — was addressed in #366. The current behavior shares the same surface but has a different root cause: it's not a styling drift, it's missing/lost broadcasts for *some* PMs in a batch.

## Hypothesis

The leak is correlated with **batch tool calls** — when one LLM response contains multiple `tool_use` blocks, the system creates one `PendingMessage` per tool, each is executed, and each is then promoted by `DrainJob`. The fact that *some but not all* PMs leak (and only when there are multiple in the same round) points at a per-broadcast issue rather than a per-session issue.

Two candidate failure modes worth examining:

1. **Some `broadcast_removed` events don't reach the TUI.** `PendingMessage#after_destroy_commit :broadcast_removed` fires `ActionCable.server.broadcast` directly (`app/models/pending_message.rb:121, 422`). If a render/decorator path raises during `broadcast_payload` for a specific PM, the broadcast for that one PM is lost while siblings go through. The TUI never learns to remove the stale entry.
2. **The TUI's removal handling skips when the matching create broadcast was missed.** Race: PM is created → broadcast fires → TUI subscriber not yet attached → `broadcast_removed` arrives later for an entry the TUI never saw create. Less likely given persistent ActionCable subscription, but possible during reconnects.

The "only some, only in batches" pattern strongly suggests **(1)** — a per-PM render path that fails for certain tool_response shapes.

## Research approach

Wire up debug logging (dev environment only) on the broadcast path to find the answer:

- Log every `PendingMessage#broadcast_created` and `#broadcast_removed` invocation with PM id, message_type, source_name, and tool_use_id.
- Log every `MessageBroadcaster#emit` invocation with message id and tool_use_id.
- On the TUI side, log every received broadcast with action (create/remove/update) and id.

Then reproduce with a session that triggers a batch tool round (e.g. spawn 5 sub-agents in parallel — exactly what triggered today's session). Compare:

- N PMs created → N `broadcast_created` logged → N broadcasts received by TUI?
[...]

PR #486, opened 2026-04-30
#481 chore: add diagnostic logging to broadcast/drain pipeline

hoblin:
## Summary

Investigative logging for #481 (TUI pending tool responses leak: some entries persist until restart). Adds a paired server + TUI logging stream so a leak round can be replayed post-hoc.

- **Server side** — new `BroadcastDiagnostics.logger` (Melete/Mneme module-logger pattern) writes to `log/broadcast.log` in dev only:
  - `PendingMessage#broadcast_created` / `#broadcast_removed` — every emit, with PM id, message_type, source_name, tool_use_id, session_id, view_mode
  - `MessageBroadcaster#emit` — every Message broadcast, with id + tool_use_id + action
  - `DrainJob#perform` — when `start_processing!` refuses, log state + `tool_round_complete?` + mailbox counts
  - `DrainJob#drain_mailbox` — round membership (tool_response PM ids + tool_use_ids), promoted count
- **TUI side** — new `TUI::BroadcastLogger` (PerformanceLogger flag-gated pattern), opt-in via `--broadcast-debug`, writes to `log/tui_broadcast.log`:
  - Every received WebSocket message logged with `action`, `type`, `id`, `pending_message_id`, `tool_use_id`
  - For `pending_message_removed`: logs whether `MessageStore#remove_pending` actually matched a stored entry (the silent-miss case is the leak's smoking gun)
- TOML template gets a new `[broadcast] log_path` setting; existing users will need `anima migrate` to pick it up before passing `--broadcast-debug`.

## Why this approach

Research ruled out the original "decorator raise eats the broadcast" framing — `drain_mailbox` runs each `promote!` in its own transaction, so a per-PM render failure can't take down siblings. Research also ruled out the comment's revised hypothesis that a `from_melete_goal` unknown-tool error breaks `Session#tool_round_complete?` — unknown-tool PMs have `message_type: \"tool_response\"` and the correct `tool_use_id`, so the round-completion query counts them correctly. The remaining failure modes (broadcast lost in transit, TUI subscription race on reconnect, `remove_pending` silent miss because the create broadcast was missed) all need the *paired* server-emit / TUI-receipt log to diagnose, which is what this PR delivers.

## How to use

```bash
# Server (always on in dev)
tail -f log/broadcast.log

# TUI side (opt-in)
./exe/anima tui --host localhost:42135 --broadcast-debug
tail -f log/tui_broadcast.log
```

Reproduce a batch tool round (e.g. spawn multiple sub-agents in parallel), wait for the leak, then compare:
- N PMs created → N `broadcast_created` server lines → N `recv pending_message_created` TUI lines?
- N PMs destroyed → N `broadcast_removed` server lines → N `recv pending_message_removed matched=true` TUI lines?
- N Messages created → N `MessageBroadcaster ... broadcast` server lines → matching `recv action=create type=tool_response` TUI lines?

Any mismatch identifies the leak point.

## Test plan

- [x] `bundle exec rspec spec/models/pending_message_spec.rb spec/jobs/drain_job_spec.rb spec/lib/events/subscribers/message_broadcaster_spec.rb spec/lib/tui/` — 748 examples pass
- [x] `bundle exec rspec spec/lib/anima/cli_spec.rb` — updated to assert new `broadcast_debug` kwarg, new spec for `--broadcast-debug` flag
- [x] `bundle exec standardrb` — clean
- [ ] Manual smoke: launch dev brain + TUI with `--broadcast-debug`, trigger a multi-tool round, verify both log files populate; record findings on #481
- [ ] After root cause identified: open follow-up implementation issue and reference from #481 (per AC)

## Breaking changes

- TUI users who already have `~/.anima/tui.toml` will need to run `anima migrate` to pick up the new `[broadcast] log_path` setting before passing `--broadcast-debug`. Users who never pass the flag are unaffected (the setting is only read when the flag is on).

Closes part of #481 (the diagnostic-logging acceptance criterion). Root-cause identification + follow-up implementation issue are the remaining ACs and will be handled after the diagnostic logs let us reproduce.

### DIFF
--- a/app/jobs/drain_job.rb
+++ b/app/jobs/drain_job.rb
@@ -48,7 +48,15 @@ class DrainJob < ApplicationJob
   # @param session_id [Integer]
   def perform(session_id)
     @session = Session.find(session_id)
-    return unless @session.start_processing!
+    unless @session.start_processing!
+      diagnostics_log.info(
+        "session=#{session_id} DrainJob — start_processing! refused: " \
+        "state=#{@session.aasm_state} round_complete?=#{@session.tool_round_complete?} " \
+        "mailbox_active=#{@session.pending_messages.active.count} " \
+        "mailbox_background=#{@session.pending_messages.background.count}"
+      )
+      return
+    end
 
     drained = drain_mailbox
     return @session.response_complete! if drained.zero?
@@ -83,6 +91,19 @@ class DrainJob < ApplicationJob
     @active_pm = @session.pending_messages.active.where.not(message_type: "tool_response").order(:created_at).first
 
     promoted = tool_responses.size + (@active_pm ? 1 : 0)
+
+    diagnostics_log.info(
+      "session=#{@session.id} drain_mailbox — tool_responses=#{tool_responses.size} " \
+      "active_pm=#{@active_pm&.id || "nil"}#{"(#{@active_pm.message_type})" if @active_pm} " \
+      "background=#{@session.pending_messages.background.count} promoted=#{promoted}"
+    )
+    if tool_responses.any?
+      diagnostics_log.debug(
+        "  tool_response PMs: " +
+          tool_responses.map { |pm| "PM##{pm.id}(uid=#{pm.tool_use_id} src=#{pm.source_name})" }.join(", ")
+      )
+    end
+
     return 0 if promoted.zero?
 
     tool_responses.each(&:promote!)
@@ -166,4 +187,6 @@ class DrainJob < ApplicationJob
   def shell_session
     @shell_session ||= ShellSession.for_session(@session)
   end
+
+  def diagnostics_log = BroadcastDiagnostics.logger
 end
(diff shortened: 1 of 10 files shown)


hoblin:
Closing — over-engineered for the goal. The acceptance criterion is to identify the broadcast leak's root cause once; shipping a permanent module + TOML setting + CLI flag + TUI logger class for a one-shot investigation is dead weight in the gem.

Replacing with inline `Rails.logger.debug` calls on the server-side broadcast/drain path (no new files, no shipped infrastructure). For first-pass diagnosis, server logs are sufficient — every `broadcast_removed` emit will show up alongside whatever the TUI is displaying, and any stuck entry whose PM id has no matching emit line points at a server-side leak; whose PM id has a matching emit line points at TUI-side application miss. TUI-side instrumentation can be added as a follow-up if first-pass diagnosis lands at #2.

PR #487, opened 2026-04-30
#480 fix: silently strip from_* tool calls before dispatch

### LINKED ISSUE #480: Research spike: source of `from_*` tool calls in the main loop

hoblin:
## Summary

Implements the fix for #480 based on the research-spike findings recorded in [the comment on the issue](https://github.com/hoblin/anima/issues/480#issuecomment-4350731469).

The spike (PR #485) confirmed that spurious `from_*` tool calls — `from_shell-runner`, `from_zero-width-sleuth`, `from_melete_goal`, etc. — are **LLM hallucinations**, not internal leaks. The model emits them as legitimate `tool_use` blocks in the raw Anthropic response, complete with fabricated input schemas, despite the `from_*` naming convention from PR #445 + system-prompt sister block + a Level 2 memory explicitly documenting the anti-pattern. Three reproductions in ~3 minutes of a single deliberately-constructed session.

## What this PR does

A small pure filter — `Aoide::PhantomCallFilter.call(hash) → hash` — drops every block where `type == "tool_use" && name.start_with?("from_")` from `response["content"]` before the rest of the main loop sees it. The filter has no AR, no events, no I/O — pure hash transformation, easy to spec in isolation.

It runs at the top of `Events::Subscribers::LLMResponseHandler#emit`, **immediately after `log_raw_response`** so the dev-only diagnostic log from PR #485 keeps observing the *pre-filter* API response. This matters: we want to keep measuring hallucination rate after the fix lands, and decide later whether a feedback loop (synthesised tool_result, Mneme nudge) is worth adding.

After the strip, the existing handler logic carries the rest of the work unchanged:

- `normalize_tool_uses` operates on the filtered content → no `from_*` enters dispatch.
- `extract_text` is unaffected by the strip → text always survives.
- The `if tool_uses.any? … elsif session.may_response_complete? …` branch handles the empty case naturally → `from_*`-only responses land at clean `response_complete!`.

## Three edge cases the user explicitly asked to cover

| Response shape | Behaviour |
| --- | --- |
| Only `from_*` tool calls | No dispatch, no failed `tool_result`, session → `idle` (clean end of turn) |
| `from_*` + legitimate tool calls | Only legitimate calls dispatch; session → `executing`; one `tool_call` message persisted |
| Text + `from_*` tool calls | Text persisted as `agent_message`; no dispatch; session → `idle` |

All three are asserted in `spec/lib/events/subscribers/llm_response_handler_spec.rb`. The filter itself has 13 unit specs against fixture payloads (including symbol-keyed content, missing content key, non-array content, malformed blocks).

## Diagnostic continuity

`log/aoide.log` keeps logging the raw API response and raw `tool_use` blocks pre-filter. Post-fix, `dispatching tool=from_*` lines should disappear, but `from_*` blocks may still appear in `raw tool_use blocks:` — that's the signal that the model is still hallucinating. If frequency stays high we'll know without re-instrumenting.

## Test plan

- [x] `bundle exec rspec spec/lib/aoide/phantom_call_filter_spec.rb` — 13 examples, 0 failures
- [x] `bundle exec rspec spec/lib/events/subscribers/llm_response_handler_spec.rb` — 16 examples (3 new edge cases), 0 failures
- [x] `bundle exec rspec spec/jobs/drain_job_spec.rb` — 11 examples, 0 failures (sanity on upstream emitter)
- [x] `bundle exec standardrb` — clean on changed files
- [x] `bundle exec reek lib/aoide/phantom_call_filter.rb lib/events/subscribers/llm_response_handler.rb` — 2 pre-existing FeatureEnvy warnings remain, none introduced
- [ ] Smoke test on dev brain — restart on this branch, reproduce the parallel-subagent scenario, confirm `dispatching tool=from_*` lines no longer appear and no `UnknownToolError` lands in the conversation. (Doing right after this PR opens.)

Closes #480

### DIFF
--- a/lib/aoide/phantom_call_filter.rb
+++ b/lib/aoide/phantom_call_filter.rb
@@ -33,7 +33,8 @@ module Aoide
       filtered = content.reject { |block| phantom_tool_use?(block) }
       return response if filtered.size == content.size
 
-      response.merge("content" => filtered)
+      key = response.key?("content") ? "content" : :content
+      response.merge(key => filtered)
     end
 
     def self.phantom_tool_use?(block)
--- a/lib/aoide/phantom_call_filter.rb
+++ b/lib/aoide/phantom_call_filter.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+module Aoide
+  # Strips +from_*+ tool_use blocks from a raw Anthropic response before
+  # the rest of the main loop sees them.
+  #
+  # The +from_*+ prefix is reserved for messages delivered *to* the
+  # agent — phantom tool_call/tool_response pairs assembled by
+  # +PendingMessage#promote!+ to surface sister-muse and sub-agent
+  # output as conversation turns. They are never registered as
+  # callable tools, so when the model hallucinates a +from_*+ tool_use
+  # block (typically while waiting for a sub-agent's push delivery),
+  # +Tools::Registry+ raises +UnknownToolError+, the failure is
+  # persisted, and tokens are wasted on a round-trip the model
+  # already had to be told not to make. This filter drops those
+  # blocks at the entry point of the response handler so they never
+  # reach dispatch.
+  #
+  # Pure: takes a hash, returns a hash. No I/O, no AR, no events.
+  module PhantomCallFilter
+    PHANTOM_PREFIX = "from_"
+
+    # Returns +response+ with every +from_*+ tool_use block removed
+    # from its +content+ array. If no such block is present, returns
+    # +response+ unchanged (same object, same identity).
+    #
+    # @param response [Hash] raw Anthropic response payload
+    # @return [Hash] sanitized response
+    def self.call(response)
+      content = response["content"] || response[:content]
+      return response unless content.is_a?(Array)
+
+      filtered = content.reject { |block| phantom_tool_use?(block) }
+      return response if filtered.size == content.size
+
+      response.merge("content" => filtered)
+    end
+
+    def self.phantom_tool_use?(block)
+      return false unless block.is_a?(Hash)
+
+      type = block["type"] || block[:type]
+      name = block["name"] || block[:name]
+      type == "tool_use" && name.is_a?(String) && name.start_with?(PHANTOM_PREFIX)
+    end
+    private_class_method :phantom_tool_use?
+  end
+end
--- a/lib/events/subscribers/llm_response_handler.rb
+++ b/lib/events/subscribers/llm_response_handler.rb
@@ -28,6 +28,7 @@ module Events
         api_metrics = payload[:api_metrics]
 
         log_raw_response(session, response)
+        response = Aoide::PhantomCallFilter.call(response)
 
         tool_uses = normalize_tool_uses(response)
         text = extract_text(response)
--- a/spec/lib/aoide/phantom_call_filter_spec.rb
+++ b/spec/lib/aoide/phantom_call_filter_spec.rb
@@ -1,0 +1,132 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Aoide::PhantomCallFilter do
+  describe ".call" do
+    it "returns the response unchanged when content has no tool_use blocks" do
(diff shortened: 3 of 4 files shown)


bonk-moltbot (approved):
🐭✅ From spike to fix — measure, then cut.

**PhantomCallFilter** is exactly the right shape: pure function, hash in → hash out, no AR, no events, no I/O. Immutable — `.merge` instead of mutating `content` in-place, original response untouched. 13 unit specs cover every edge: symbol keys, nil content, bare prefix, non-hash blocks, multiple phantoms, mixed real+phantom.

**Placement is surgical:** after `log_raw_response` (diagnostic data preserved pre-filter) but before `normalize_tool_uses` (no `from_*` enters dispatch). The logging spec updated accordingly — raw blocks still show `from_*`, but no `dispatching tool=from_*` lines anymore.

**Three edge cases from the ticket explicitly covered in integration specs:**
- `from_*`-only → clean idle, no failed tool_result, no wasted tokens
- `from_*` + legitimate → only legitimate dispatches, one tool_call persisted
- text + `from_*` → text kept as agent_message, session idles cleanly

The filter keeps hallucination rate observable (raw blocks in aoide.log) while eliminating the symptom (UnknownToolError, wasted round-trips). Future feedback loop (synthesised tool_result, Mneme nudge) remains a clean extension point.

